PR #8, opened 2023-03-27
Update real-world test certificates

complexspaces:
These certificates expired a while ago, but CI hasn't ran since so it wasn't an issue. This PR updates both of the real-world certificates so that the versions used in tests are no longer expired.


src/tests/verification_real_world/update_valid_1_cert.bash
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash
 set -euo pipefail
 
-echo 'this script only update 1password_com_valid_1.crt.'
+echo 'This script only update 1password_com_valid_1.crt'
```
cpu, line 4: Another grammar nit while you're here:

```suggestion
echo 'This script only updates 1password_com_valid_1.crt'
```

PR #9, opened 2023-03-27
update rustls v0.20 -> v0.21

cpu:
## Description 

This branch updates `rustls-platform-verifier` to use the recently released rustls 0.21.0, both as a direct dependency and through upgrades to relevant transitive deps [reqwest](https://github.com/seanmonstar/reqwest/pull/1791), [hyper-rustls](https://github.com/rustls/hyper-rustls/pull/199), and [tokio-rustls](https://github.com/tokio-rs/tls/pull/137)).

### Details

Most notable for this codebase, the `rustls::Error::InvalidCertificateData` has been removed and replaced
by `InvalidCertificate` and a number of `CertificateError` sub-variants.

Now that the upstream `InvalidCertificate` error offers a way to specify what went wrong with more granularity we're able to trim all of the `error_messages` consts, choosing instead to return the more specific upstream error types. For the case where invalid extensions are detected we define our own error type to use consistently across verifiers.

In all other circumstances where we want to return an `InvalidCertificate` error with some platform specific error message we
use the `InvalidCertificate(CertificateError::Other)` variant with the error message, re-purposing the `invalid_certificate` helper for constructing an `Arc` over a `Box` with the error message.

We also have to take some special care when asserting the equality of errors, handling the case where looking at a `CertificateError::Other` specially, since it can box a dyn error that can't be compared directly without downcasting the error to a concrete type.

### Remaining work

- [X] wait for https://github.com/tokio-rs/tls/pull/137 to be merged
- [x] wait for tokio-rustls v0.24.0 to be published to crates.io
- [x] remove Cargo.toml patch for tokio-rustls
- [x] wait for https://github.com/rustls/hyper-rustls/pull/199 to be merged
- [x] wait for hyper-rustls 0.24.0 to be published to crates.io
- [x] remove Cargo.toml patch for hyper-rustls
- [x] wait for https://github.com/seanmonstar/reqwest/pull/1791 to be merged
- [x] wait for reqwest 0.11.18 to be published to crates.io
- [x] remove Cargo.toml patch for reqwest
- [X] get this PR approved.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -37,7 +37,7 @@ cert-logging = ["base64"]
 docsrs = ["jni", "once_cell"]
 
 [dependencies]
-rustls = { version = "0.20", features = ["dangerous_configuration", "tls12", "logging"] }
+rustls = { version = "0.21", features = ["dangerous_configuration", "tls12", "logging"] }
 log = { version = "0.4" }
 base64 = { version = "0.13", optional = true } # Only used when the `cert-logging` feature is enabled.
 jni = { version = "0.19", default-features = false, optional = true } # Only used during doc generation
@@ -46,17 +46,18 @@ once_cell = { version = "1.9", optional = true } # Only used during doc generati
 [target.'cfg(target_os = "linux")'.dependencies]
 rustls-native-certs = "0.6"
 once_cell = "1.9"
-webpki-roots = "0.22" # Augmentation to `openssl-probe` due to inconsistencies on Linux distros.
+webpki-roots = "0.23" # Augmentation to `openssl-probe` due to inconsistencies on Linux distros.
+webpki = { package = "rustls-webpki", version = "0.100.0", features = ["alloc", "std"] }
 
 [target.'cfg(target_os = "android")'.dependencies]
 jni = { version = "0.19", default-features = false }
-webpki = "0.22"
+webpki = { package = "rustls-webpki", version = "0.100.0", features = ["alloc", "std"] }
 once_cell = "1.9"
 android_logger = { version = "0.11", optional = true } # Only used during testing.
 
 [target.'cfg(target_arch = "wasm32")'.dependencies]
 once_cell = "1.9"
-webpki-roots = "0.22"
+webpki-roots = "0.23"
 
 [target.'cfg(any(target_os = "macos", target_os = "ios"))'.dependencies]
 core-foundation = "0.9"
@@ -69,7 +70,7 @@ winapi = { version = "0.3", features = ["wincrypt", "winerror"] }
 
 [dev-dependencies]
 tokio = { version = "1.5", features = ["macros", "rt-multi-thread"] }
-reqwest = { version = "0.11", default-features = false, features = ["rustls-tls-manual-roots"] }
+reqwest = { version = "0.11.18", default-features = false, features = ["rustls-tls-manual-roots"] }
 
 [package.metadata.docs.rs]
 rustdoc-args = ["--cfg", "docsrs"]
--- a/src/tests/mod.rs
+++ b/src/tests/mod.rs
@@ -1,11 +1,15 @@
 #[cfg(feature = "ffi-testing")]
 pub mod ffi;
 
+use std::error::Error as StdError;
+
 mod verification_real_world;
 
 mod verification_mock;
 
-struct TestCase<'a> {
+use rustls::{CertificateError, Error as TlsError, Error::InvalidCertificate};
+
+struct TestCase<'a, E: StdError> {
     /// The name of the server we're connecting to.
     pub reference_id: &'a str,
 
@@ -15,5 +19,30 @@ struct TestCase<'a> {
     /// The stapled OCSP response given to us by Rustls, if any.
     pub stapled_ocsp: Option<&'a [u8]>,
 
-    pub expected_result: Result<(), rustls::Error>,
+    pub expected_result: Result<(), TlsError>,
+
+    /// An error that should be present inside an expected `CertificateError::Other` variant.
+    ///
+    /// Set this if the error being tested uses `CertificateError::Other` and not statically known
+    /// variants in [TlsError]
+    #[allow(dead_code)]
+    pub other_error: Option<E>,
+}
+
+pub fn assert_cert_error_eq<E: StdError + PartialEq + 'static>(
+    result: &Result<(), TlsError>,
+    expected: &Result<(), TlsError>,
+    expected_err: Option<&E>,
+) {
+    // If the expected error is an "Other" CertificateError we can't directly assert equality, we rely
+    // on the test caller to provide the correct value to compare.
+    if let Err(InvalidCertificate(CertificateError::Other(err))) = &expected {
+        let expected_err = expected_err.expect("error not provided for `Other` case handling");
+        let err: &E = err
+            .downcast_ref()
+            .expect("incorrect `Other` inner error kind");
+        assert_eq!(err, expected_err);
+    } else {
+        assert_eq!(result, expected);
+    }
 }
--- a/src/tests/verification_mock/mod.rs
+++ b/src/tests/verification_mock/mod.rs
@@ -21,10 +21,12 @@
 ))]
 
 use super::TestCase;
-use crate::verification::{error_messages, Verifier};
-use rustls::{client::ServerCertVerifier, Error as TlsError};
+use crate::tests::assert_cert_error_eq;
+use crate::verification::{EkuError, Verifier};
+use rustls::{client::ServerCertVerifier, CertificateError, Error as TlsError};
 use std::convert::TryFrom;
 use std::net::IpAddr;
+use std::sync::Arc;
 
 macro_rules! mock_root_test_cases {
     { $( $name:ident [ $target:meta ] => $test_case:expr ),+ , } => {
@@ -58,6 +60,12 @@ macro_rules! mock_root_test_cases {
                 test_with_mock_root(&$test_case);
             }
         )+
+    };
+}
+
+macro_rules! no_error {
+    () => {
+        None::<std::convert::Infallible>
     };
 }
 
@@ -96,9 +104,9 @@ pub(super) fn verification_without_mock_root() {
 
     assert_eq!(
         result.map(|_| ()),
-        Err(TlsError::InvalidCertificateData(String::from(
-            error_messages::UNKNOWN_CERT
-        )))
+        Err(TlsError::InvalidCertificate(
+            CertificateError::UnknownIssuer
+        ))
     );
 }
 
@@ -117,36 +125,42 @@ mock_root_test_cases! {
         chain: &[ROOT1_INT1_EXAMPLE_COM_GOOD, ROOT1_INT1],
         stapled_ocsp: None,
         expected_result: Ok(()),
+        other_error: no_error!(),
     },
     valid_no_stapling_ipv4 [ any(windows, target_os = "macos") ] => TestCase {
         reference_id: LOCALHOST_IPV4,
         chain: &[ROOT1_INT1_LOCALHOST_IPV4_GOOD, ROOT1_INT1],
         stapled_ocsp: None,
         expected_result: Ok(()),
+        other_error: no_error!(),
     },
     valid_no_stapling_ipv6 [ any(windows, target_os = "macos") ] => TestCase {
         reference_id: LOCALHOST_IPV6,
         chain: &[ROOT1_INT1_LOCALHOST_IPV6_GOOD, ROOT1_INT1],
         stapled_ocsp: None,
         expected_result: Ok(()),
+        other_error: no_error!(),
     },
     valid_stapled_good_dns [ any(windows, target_os = "android", target_os = "macos", target_os = "linux") ] => TestCase {
         reference_id: EXAMPLE_COM,
         chain: &[ROOT1_INT1_EXAMPLE_COM_GOOD, ROOT1_INT1],
         stapled_ocsp: Some(include_bytes!("root1-int1-ee_example.com-good.ocsp")),
         expected_result: Ok(()),
+        other_error: no_error!(),
     },
     valid_stapled_good_ipv4 [ any(windows, target_os = "macos") ] => TestCase {
         reference_id: LOCALHOST_IPV4,
         chain: &[ROOT1_INT1_LOCALHOST_IPV4_GOOD, ROOT1_INT1],
         stapled_ocsp: Some(include_bytes!("root1-int1-ee_127.0.0.1-good.ocsp")),
         expected_result: Ok(()),
+        other_error: no_error!(),
     },
     valid_stapled_good_ipv6 [ any(windows, target_os = "macos") ] => TestCase {
         reference_id: LOCALHOST_IPV6,
         chain: &[ROOT1_INT1_LOCALHOST_IPV6_GOOD, ROOT1_INT1],
         stapled_ocsp: Some(include_bytes!("root1-int1-ee_1-good.ocsp")),
         expected_result: Ok(()),
+        other_error: no_error!(),
     },
     // Uses a separate certificate from the one used in the "good" case to deal
     // with operating systems with validation data caches (e.g. Windows).
@@ -154,19 +168,22 @@ mock_root_test_cases! {
         reference_id: EXAMPLE_COM,
         chain: &[include_bytes!("root1-int1-ee_example.com-revoked.crt"), ROOT1_INT1],
         stapled_ocsp: Some(include_bytes!("root1-int1-ee_example.com-revoked.ocsp")),
-        expected_result: Err(TlsError::InvalidCertificateData(String::from(error_messages::REVOKED))),
+        expected_result: Err(TlsError::InvalidCertificate(CertificateError::Revoked)),
+        other_error: no_error!(),
     },
     stapled_revoked_ipv4 [ any(windows, target_os = "macos") ] => TestCase {
         reference_id: LOCALHOST_IPV4,
         chain: &[include_bytes!("root1-int1-ee_127.0.0.1-revoked.crt"), ROOT1_INT1],
         stapled_ocsp: Some(include_bytes!("root1-int1-ee_127.0.0.1-revoked.ocsp")),
-        expected_result: Err(TlsError::InvalidCertificateData(String::from(error_messages::REVOKED))),
+        expected_result: Err(TlsError::InvalidCertificate(CertificateError::Revoked)),
+        other_error: no_error!(),
     },
     stapled_revoked_ipv6 [ any(windows, target_os = "macos") ] => TestCase {
         reference_id: LOCALHOST_IPV6,
         chain: &[include_bytes!("root1-int1-ee_1-revoked.crt"), ROOT1_INT1],
         stapled_ocsp: Some(include_bytes!("root1-int1-ee_1-revoked.ocsp")),
-        expected_result: Err(TlsError::InvalidCertificateData(String::from(error_messages::REVOKED))),
+        expected_result: Err(TlsError::InvalidCertificate(CertificateError::Revoked)),
+        other_error: no_error!(),
     },
     // Validation fails with no intermediate (that can't be fetched
     // with AIA because there's no AIA issuer field in the certificate).
@@ -177,60 +194,72 @@ mock_root_test_cases! {
         reference_id: EXAMPLE_COM,
         chain: &[ROOT1_INT1_EXAMPLE_COM_GOOD],
         stapled_ocsp: None,
-        expected_result: Err(TlsError::InvalidCertificateData(String::from(error_messages::UNKNOWN_CERT))),
+        expected_result: Err(TlsError::InvalidCertificate(CertificateError::UnknownIssuer)),
+        other_error: no_error!(),
     },
     ee_only_ipv4 [ any(windows, target_os = "macos") ] => TestCase {
         reference_id: LOCALHOST_IPV4,
         chain: &[ROOT1_INT1_LOCALHOST_IPV4_GOOD],
         stapled_ocsp: None,
-        expected_result: Err(TlsError::InvalidCertificateData(String::from(error_messages::UNKNOWN_CERT))),
+        expected_result: Err(TlsError::InvalidCertificate(CertificateError::UnknownIssuer)),
+        other_error: no_error!(),
     },
     ee_only_ipv6 [ any(windows, target_os = "macos") ] => TestCase {
         reference_id: LOCALHOST_IPV6,
         chain: &[ROOT1_INT1_LOCALHOST_IPV6_GOOD],
         stapled_ocsp: None,
-        expected_result: Err(TlsError::InvalidCertificateData(String::from(error_messages::UNKNOWN_CERT))),
+        expected_result: Err(TlsError::InvalidCertificate(CertificateError::UnknownIssuer)),
+        other_error: no_error!(),
     },
     // Validation fails when the certificate isn't valid for the reference ID.
     domain_mismatch_dns [ any(windows, target_os = "android", target_os = "macos", target_os = "linux") ] => TestCase {
         reference_id: "example.org",
         chain: &[ROOT1_INT1_EXAMPLE_COM_GOOD, ROOT1_INT1],
         stapled_ocsp: None,
-        expected_result: Err(TlsError::InvalidCertificateData(String::from(error_messages::WRONG_NAME))),
+        expected_result: Err(TlsError::InvalidCertificate(CertificateError::NotValidForName)),
+        other_error: no_error!(),
     },
     domain_mismatch_ipv4 [ any(windows, target_os = "macos") ] => TestCase {
         reference_id: "198.168.0.1",
         chain: &[ROOT1_INT1_LOCALHOST_IPV4_GOOD, ROOT1_INT1],
         stapled_ocsp: None,
-        expected_result: Err(TlsError::InvalidCertificateData(String::from(error_messages::WRONG_NAME))),
+        expected_result: Err(TlsError::InvalidCertificate(CertificateError::NotValidForName)),
+        other_error: no_error!(),
     },
     domain_mismatch_ipv6 [ any(windows, target_os = "macos") ] => TestCase {
         reference_id: "::ffff:c6a8:1",
         chain: &[ROOT1_INT1_LOCALHOST_IPV6_GOOD, ROOT1_INT1],
         stapled_ocsp: None,
-        expected_result: Err(TlsError::InvalidCertificateData(String::from(error_messages::WRONG_NAME))),
+        expected_result: Err(TlsError::InvalidCertificate(CertificateError::NotValidForName)),
+        other_error: no_error!(),
     },
     wrong_eku_dns [ any(windows, target_os = "android", target_os = "macos", target_os = "linux") ] => TestCase {
         reference_id: EXAMPLE_COM,
         chain: &[include_bytes!("root1-int1-ee_example.com-wrong_eku.crt"), ROOT1_INT1],
         stapled_ocsp: None,
-        expected_result: Err(TlsError::InvalidCertificateData(String::from(error_messages::INVALID_EXTENSIONS))),
+        expected_result: Err(TlsError::InvalidCertificate(
+            CertificateError::Other(Arc::from(EkuError)))),
+        other_error: Some(EkuError),
     },
     wrong_eku_ipv4 [ any(windows, target_os = "macos") ] => TestCase {
         reference_id: LOCALHOST_IPV4,
         chain: &[include_bytes!("root1-int1-ee_127.0.0.1-wrong_eku.crt"), ROOT1_INT1],
         stapled_ocsp: None,
-        expected_result: Err(TlsError::InvalidCertificateData(String::from(error_messages::INVALID_EXTENSIONS))),
+        expected_result: Err(TlsError::InvalidCertificate(
+            CertificateError::Other(Arc::from(EkuError)))),
+        other_error: Some(EkuError),
     },
     wrong_eku_ipv6 [ any(windows, target_os = "macos") ] => TestCase {
         reference_id: LOCALHOST_IPV6,
         chain: &[include_bytes!("root1-int1-ee_1-wrong_eku.crt"), ROOT1_INT1],
         stapled_ocsp: None,
-        expected_result: Err(TlsError::InvalidCertificateData(String::from(error_messages::INVALID_EXTENSIONS))),
+        expected_result: Err(TlsError::InvalidCertificate(
+            CertificateError::Other(Arc::from(EkuError)))),
+        other_error: Some(EkuError),
     },
 }
 
-fn test_with_mock_root(test_case: &TestCase) {
+fn test_with_mock_root<E: std::error::Error + PartialEq + 'static>(test_case: &TestCase<E>) {
     log::info!("verifying {:?}", test_case.expected_result);
 
     let verifier = Verifier::new_with_fake_root(ROOT1); // TODO: time
@@ -264,6 +293,11 @@ fn test_with_mock_root(test_case: &TestCase) {
         test_case.stapled_ocsp.unwrap_or(&[]),
         std::time::SystemTime::now(),
     );
-    assert_eq!(result.map(|_| ()), test_case.expected_result);
+
+    assert_cert_error_eq(
+        &result.map(|_| ()),
+        &test_case.expected_result,
+        test_case.other_error.as_ref(),
+    );
     // TODO: get into specifics of errors returned when it fails.
 }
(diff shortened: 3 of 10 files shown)


src/tests/verification_mock/mod.rs
```diff
@@ -264,6 +265,7 @@ fn test_with_mock_root(test_case: &TestCase) {
         test_case.stapled_ocsp.unwrap_or(&[]),
         std::time::SystemTime::now(),
     );
-    assert_eq!(result.map(|_| ()), test_case.expected_result);
+    let _expected = &test_case.expected_result;
+    assert!(matches!(result.map(|_| ()), _expected));
```
cpu, line 269: This is the first time I've used the `matches!` macro and it's a little peculiar. 

Leaving the expected value in-line produces a syntax error about an unexpected `.`, similarly using `expected` not named as `_expected` raises an unused variable warning even though it clearly is used in the match expression.

This is my clumsy way to make it work. Suggestions for better approaches welcome!

cpu: Update: this is just plain wrong :sweat_smile: Looking into it.

cpu:
> cpu force-pushed the cpu-rustls-0.21.0-prep branch from 6d613c9 to 0cdc9d2 

Since Rustls [restored `PartialEq`](https://github.com/rustls/rustls/pull/1252) on the error type I was able to drop the debug assertion match helper. We're back to using `assert_eq!` with a small helper for coarsely matching the `CertificateError::Other` case. PR description updated accordingly.

src/tests/verification_mock/mod.rs
```diff
@@ -94,12 +99,13 @@ pub(super) fn verification_without_mock_root() {
...
-            error_messages::UNKNOWN_CERT
-        )))
-    );
+    let res = result.map(|_| ());
+    assert!(matches!(
+        res,
+        Err(TlsError::InvalidCertificate(
+            CertificateError::UnknownIssuer
+        ))
+    ));
```
complexspaces, line 108: With the return of `PartialEq`, this change can be reverted right?

src/verification/mod.rs
```diff
@@ -58,10 +41,13 @@ fn unsupported_server_name() -> rustls::Error {
...
 }
 
-// Certificate data error shorthand
+// Unknown certificate error shorthand. Used when we need to construct an "Other" certificate
+// error with a platform specific error message.
 #[cfg(any(windows, target_os = "android", target_os = "macos", target_os = "ios"))]
 fn invalid_certificate(reason: impl Into<String>) -> rustls::Error {
-    rustls::Error::InvalidCertificateData(reason.into())
+    rustls::Error::InvalidCertificate(rustls::CertificateError::Other(std::sync::Arc::from(
+        Box::from(reason.into()),
```
complexspaces, line 49: This might be reducible to `reason.into()` given [the implementation](https://doc.rust-lang.org/std/error/trait.Error.html#impl-From%3CString%3E-for-Box%3Cdyn%20Error%20+%20%27static%2C%20Global%3E) of `From<String> for Box<Error>`.

src/verification/windows.rs
```diff
@@ -575,12 +573,11 @@ impl Verifier {
...
         Err(match win_error {
             CERT_E_CN_NO_MATCH | CERT_E_INVALID_NAME => {
-                invalid_certificate(error_messages::WRONG_NAME)
+                InvalidCertificate(CertificateError::NotValidForName)
             }
-            CRYPT_E_REVOKED => invalid_certificate(error_messages::REVOKED),
+            CRYPT_E_REVOKED => InvalidCertificate(CertificateError::Revoked),
             error_num => {
                 let err = std::io::Error::from_raw_os_error(error_num);
-                // The included error message has both the description and raw OS error code.
```
complexspaces, line 583: Nit: I'm not sure why this was removed.

src/tests/mod.rs
```diff
@@ -15,5 +17,18 @@ struct TestCase<'a> {
...
+}
+
+pub fn assert_cert_error_eq(result: &Result<(), TlsError>, expected: &Result<(), TlsError>) {
+    // If the expected error is an "Other" CertificateError we can't directly assert equality.
+    // Instead, assert that the actual error matches an "Other" CertificateError pattern.
+    if let Err(InvalidCertificate(CertificateError::Other(_))) = &expected {
+        assert!(matches!(
+            result,
+            Err(InvalidCertificate(CertificateError::Other(_)))
+        ))
```
complexspaces, line 30: With these changes, we no longer seem to be asserting that the bad EKU errors exactly match. Instead if something bugged out and we got a more generic string error, it would also pass.

What do you think about passing an optional concrete type as `expected_err: Option<impl PartialEq>`, and if its present, downcast the contents of `CertificateError::Other` and check that they are the same? 

Given that the only remaining case in tests that this is used is for the EKUs, we could probably see if we have this pattern and then pass in the expected error data based on the same `#[cfg()]` values used in the test cases themselves. Does that seem reasonable to accomplish or is the idea's complexity oversimplified?

src/tests/verification_mock/mod.rs
cpu: Good catch. I missed this in my cleanup. Will adjust.

src/verification/mod.rs
cpu: Hmm, no luck (or I'm holding it wrong):

```rust
    rustls::Error::InvalidCertificate(rustls::CertificateError::Other(std::sync::Arc::from(
        reason.into(),
    )))
```
->
```
error[E0277]: the trait bound `std::sync::Arc<dyn std::error::Error + std::marker::Send + std::marker::Sync>: std::convert::From<std::string::String>` is not satisfied
```

src/verification/windows.rs
cpu: Refactoring mistake, will restore.

src/tests/mod.rs
cpu: > What do you think about passing an optional concrete type as expected_err: Option<impl PartialEq>, and if its present, downcast the contents of CertificateError::Other and check that they are the same?

That sounds reasonable. I'll iterate on this. :+1:

cpu: I'm not sure how we can know what concrete type to downcast the error to in order to compare it to the `impl PartialEq`. Am I overlooking something?

src/verification/mod.rs
complexspaces: It looks like you were right and I was incorrect here. Its a bit unfortunate that `Other` error reporting with `String` now requires double allocation now but I don't think there's anything that can be done about it.

src/tests/mod.rs
complexspaces: This ended up being more involved then I made it out to be originally 😓. As mentioned on Discord, I decided to do away with the last `String` error in our tests and replace it with a real type that can be used instead. With that in place, I changed the tests to support carrying this optional expectation to enforce when `CertificateError::Other` is encountered. Please have a look and see if it makes sense:
```diff
diff --git a/Cargo.toml b/Cargo.toml
index da1812d..21d411b 100644
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -58,6 +58,7 @@ android_logger = { version = "0.11", optional = true } # Only used during testin
 [target.'cfg(target_arch = "wasm32")'.dependencies]
 once_cell = "1.9"
 webpki-roots = "0.22"
+webpki = { package = "rustls-webpki", version = "0.100.0", features = ["alloc", "std"] }
 
 [target.'cfg(any(target_os = "macos", target_os = "ios"))'.dependencies]
 core-foundation = "0.9"
diff --git a/src/tests/mod.rs b/src/tests/mod.rs
index a7188ae..b3148d0 100644
--- a/src/tests/mod.rs
+++ b/src/tests/mod.rs
@@ -1,13 +1,15 @@
 #[cfg(feature = "ffi-testing")]
 pub mod ffi;
 
+use std::error::Error as StdError;
+
 mod verification_real_world;
 
 mod verification_mock;
 
 use rustls::{CertificateError, Error as TlsError, Error::InvalidCertificate};
 
-struct TestCase<'a> {
+struct TestCase<'a, E: StdError> {
     /// The name of the server we're connecting to.
     pub reference_id: &'a str,
 
@@ -18,16 +20,26 @@ struct TestCase<'a> {
     pub stapled_ocsp: Option<&'a [u8]>,
 
     pub expected_result: Result<(), TlsError>,
+    /// An error that should be present inside an expected `CertificateError::Other` variant.
+    ///
+    /// Set this if the error being tested uses `CertificateError::Other` and not statically known
+    /// variants in [TlsError]
+    pub other_error: Option<E>,
 }
 
-pub fn assert_cert_error_eq(result: &Result<(), TlsError>, expected: &Result<(), TlsError>) {
-    // If the expected error is an "Other" CertificateError we can't directly assert equality.
-    // Instead, assert that the actual error matches an "Other" CertificateError pattern.
-    if let Err(InvalidCertificate(CertificateError::Other(_))) = &expected {
-        assert!(matches!(
-            result,
-            Err(InvalidCertificate(CertificateError::Other(_)))
-        ))
+pub fn assert_cert_error_eq<E: StdError + PartialEq + 'static>(
+    result: &Result<(), TlsError>,
+    expected: &Result<(), TlsError>,
+    expected_err: Option<&E>,
+) {
+    // If the expected error is an "Other" CertificateError we can't directly assert equality, we rely
+    // on the test caller to provide the correct value to compare.
+    if let Err(InvalidCertificate(CertificateError::Other(err))) = &expected {
+        let expected_err = expected_err.expect("error not provided for `Other` case handling");
+        let err: &E = err
+            .downcast_ref()
+            .expect("incorrect `Other` inner error kind");
+        assert_eq!(err, expected_err);
     } else {
         assert_eq!(result, expected);
     }
diff --git a/src/tests/verification_mock/mod.rs b/src/tests/verification_mock/mod.rs
index 79c56ba..86de1e5 100644
--- a/src/tests/verification_mock/mod.rs
+++ b/src/tests/verification_mock/mod.rs
@@ -22,15 +22,12 @@
 
 use super::TestCase;
 use crate::tests::assert_cert_error_eq;
-use crate::verification::Verifier;
+use crate::verification::{EkuError, Verifier};
 use rustls::{client::ServerCertVerifier, CertificateError, Error as TlsError};
 use std::convert::TryFrom;
 use std::net::IpAddr;
 use std::sync::Arc;
 
-#[cfg(any(windows, target_os = "android", target_os = "macos"))]
-use crate::verification::error_messages;
-
 macro_rules! mock_root_test_cases {
     { $( $name:ident [ $target:meta ] => $test_case:expr ),+ , } => {
         mock_root_test_cases!(@ $($name [ $target ] => $test_case),+,);
@@ -66,6 +63,12 @@ macro_rules! mock_root_test_cases {
     };
 }
 
+macro_rules! no_error {
+    () => {
+        None::<std::convert::Infallible>
+    };
+}
+
 const ROOT1: &[u8] = include_bytes!("root1.crt");
 const ROOT1_INT1: &[u8] = include_bytes!("root1-int1.crt");
 const ROOT1_INT1_EXAMPLE_COM_GOOD: &[u8] = include_bytes!("root1-int1-ee_example.com-good.crt");
@@ -123,36 +126,42 @@ mock_root_test_cases! {
         chain: &[ROOT1_INT1_EXAMPLE_COM_GOOD, ROOT1_INT1],
         stapled_ocsp: None,
         expected_result: Ok(()),
+        other_error: no_error!(),
     },
     valid_no_stapling_ipv4 [ any(windows, target_os = "macos") ] => TestCase {
         reference_id: LOCALHOST_IPV4,
         chain: &[ROOT1_INT1_LOCALHOST_IPV4_GOOD, ROOT1_INT1],
         stapled_ocsp: None,
         expected_result: Ok(()),
+        other_error: no_error!(),
     },
     valid_no_stapling_ipv6 [ any(windows, target_os = "macos") ] => TestCase {
         reference_id: LOCALHOST_IPV6,
         chain: &[ROOT1_INT1_LOCALHOST_IPV6_GOOD, ROOT1_INT1],
         stapled_ocsp: None,
         expected_result: Ok(()),
+        other_error: no_error!(),
     },
     valid_stapled_good_dns [ any(windows, target_os = "android", target_os = "macos", target_os = "linux") ] => TestCase {
         reference_id: EXAMPLE_COM,
         chain: &[ROOT1_INT1_EXAMPLE_COM_GOOD, ROOT1_INT1],
         stapled_ocsp: Some(include_bytes!("root1-int1-ee_example.com-good.ocsp")),
         expected_result: Ok(()),
+        other_error: no_error!(),
     },
     valid_stapled_good_ipv4 [ any(windows, target_os = "macos") ] => TestCase {
         reference_id: LOCALHOST_IPV4,
         chain: &[ROOT1_INT1_LOCALHOST_IPV4_GOOD, ROOT1_INT1],
         stapled_ocsp: Some(include_bytes!("root1-int1-ee_127.0.0.1-good.ocsp")),
         expected_result: Ok(()),
+        other_error: no_error!(),
     },
     valid_stapled_good_ipv6 [ any(windows, target_os = "macos") ] => TestCase {
         reference_id: LOCALHOST_IPV6,
         chain: &[ROOT1_INT1_LOCALHOST_IPV6_GOOD, ROOT1_INT1],
         stapled_ocsp: Some(include_bytes!("root1-int1-ee_1-good.ocsp")),
         expected_result: Ok(()),
+        other_error: no_error!(),
     },
     // Uses a separate certificate from the one used in the "good" case to deal
     // with operating systems with validation data caches (e.g. Windows).
@@ -161,18 +170,21 @@ mock_root_test_cases! {
         chain: &[include_bytes!("root1-int1-ee_example.com-revoked.crt"), ROOT1_INT1],
         stapled_ocsp: Some(include_bytes!("root1-int1-ee_example.com-revoked.ocsp")),
         expected_result: Err(TlsError::InvalidCertificate(CertificateError::Revoked)),
+        other_error: no_error!(),
     },
     stapled_revoked_ipv4 [ any(windows, target_os = "macos") ] => TestCase {
         reference_id: LOCALHOST_IPV4,
         chain: &[include_bytes!("root1-int1-ee_127.0.0.1-revoked.crt"), ROOT1_INT1],
         stapled_ocsp: Some(include_bytes!("root1-int1-ee_127.0.0.1-revoked.ocsp")),
         expected_result: Err(TlsError::InvalidCertificate(CertificateError::Revoked)),
+        other_error: no_error!(),
     },
     stapled_revoked_ipv6 [ any(windows, target_os = "macos") ] => TestCase {
         reference_id: LOCALHOST_IPV6,
         chain: &[include_bytes!("root1-int1-ee_1-revoked.crt"), ROOT1_INT1],
         stapled_ocsp: Some(include_bytes!("root1-int1-ee_1-revoked.ocsp")),
         expected_result: Err(TlsError::InvalidCertificate(CertificateError::Revoked)),
+        other_error: no_error!(),
     },
     // Validation fails with no intermediate (that can't be fetched
     // with AIA because there's no AIA issuer field in the certificate).
@@ -184,18 +196,21 @@ mock_root_test_cases! {
         chain: &[ROOT1_INT1_EXAMPLE_COM_GOOD],
         stapled_ocsp: None,
         expected_result: Err(TlsError::InvalidCertificate(CertificateError::UnknownIssuer)),
+        other_error: no_error!(),
     },
     ee_only_ipv4 [ any(windows, target_os = "macos") ] => TestCase {
         reference_id: LOCALHOST_IPV4,
         chain: &[ROOT1_INT1_LOCALHOST_IPV4_GOOD],
         stapled_ocsp: None,
         expected_result: Err(TlsError::InvalidCertificate(CertificateError::UnknownIssuer)),
+        other_error: no_error!(),
     },
     ee_only_ipv6 [ any(windows, target_os = "macos") ] => TestCase {
         reference_id: LOCALHOST_IPV6,
         chain: &[ROOT1_INT1_LOCALHOST_IPV6_GOOD],
         stapled_ocsp: None,
         expected_result: Err(TlsError::InvalidCertificate(CertificateError::UnknownIssuer)),
+        other_error: no_error!(),
     },
     // Validation fails when the certificate isn't valid for the reference ID.
     domain_mismatch_dns [ any(windows, target_os = "android", target_os = "macos", target_os = "linux") ] => TestCase {
@@ -203,50 +218,60 @@ mock_root_test_cases! {
         chain: &[ROOT1_INT1_EXAMPLE_COM_GOOD, ROOT1_INT1],
         stapled_ocsp: None,
         expected_result: Err(TlsError::InvalidCertificate(CertificateError::NotValidForName)),
+        other_error: no_error!(),
     },
     domain_mismatch_ipv4 [ any(windows, target_os = "macos") ] => TestCase {
         reference_id: "198.168.0.1",
         chain: &[ROOT1_INT1_LOCALHOST_IPV4_GOOD, ROOT1_INT1],
         stapled_ocsp: None,
         expected_result: Err(TlsError::InvalidCertificate(CertificateError::NotValidForName)),
+        other_error: no_error!(),
     },
     domain_mismatch_ipv6 [ any(windows, target_os = "macos") ] => TestCase {
         reference_id: "::ffff:c6a8:1",
         chain: &[ROOT1_INT1_LOCALHOST_IPV6_GOOD, ROOT1_INT1],
         stapled_ocsp: None,
         expected_result: Err(TlsError::InvalidCertificate(CertificateError::NotValidForName)),
+        other_error: no_error!(),
     },
     wrong_eku_dns [ any(target_os = "linux") ] => TestCase {
         reference_id: EXAMPLE_COM,
         chain: &[include_bytes!("root1-int1-ee_example.com-wrong_eku.crt"), ROOT1_INT1],
         stapled_ocsp: None,
         expected_result: Err(TlsError::InvalidCertificate(
-            CertificateError::Other(Arc::from(Box::from(webpki::Error::RequiredEkuNotFound))))),
+            CertificateError::Other(Arc::from(EkuError)))),
+        other_error: Some(EkuError)
     },
     wrong_eku_dns [ any(windows, target_os = "android", target_os = "macos") ] => TestCase {
         reference_id: EXAMPLE_COM,
         chain: &[include_bytes!("root1-int1-ee_example.com-wrong_eku.crt"), ROOT1_INT1],
         stapled_ocsp: None,
         expected_result: Err(TlsError::InvalidCertificate(
-            CertificateError::Other(Arc::from(Box::from(error_messages::INVALID_EXTENSIONS))))),
+            CertificateError::Other(Arc::from(EkuError))
+        )),
+        other_error: Some(EkuError)
     },
     wrong_eku_ipv4 [ any(windows, target_os = "macos") ] => TestCase {
         reference_id: LOCALHOST_IPV4,
         chain: &[include_bytes!("root1-int1-ee_127.0.0.1-wrong_eku.crt"), ROOT1_INT1],
         stapled_ocsp: None,
         expected_result: Err(TlsError::InvalidCertificate(
-            CertificateError::Other(Arc::from(Box::from(error_messages::INVALID_EXTENSIONS))))),
+            CertificateError::Other(Arc::from(EkuError))
+        )),
+        other_error: Some(EkuError)
     },
     wrong_eku_ipv6 [ any(windows, target_os = "macos") ] => TestCase {
         reference_id: LOCALHOST_IPV6,
         chain: &[include_bytes!("root1-int1-ee_1-wrong_eku.crt"), ROOT1_INT1],
         stapled_ocsp: None,
         expected_result: Err(TlsError::InvalidCertificate(
-            CertificateError::Other(Arc::from(Box::from(error_messages::INVALID_EXTENSIONS))))),
+            CertificateError::Other(Arc::from(EkuError))
+        )),
+        other_error: Some(EkuError)
     },
 }
 
-fn test_with_mock_root(test_case: &TestCase) {
+fn test_with_mock_root<E: std::error::Error + PartialEq + 'static>(test_case: &TestCase<E>) {
     log::info!("verifying {:?}", test_case.expected_result);
 
     let verifier = Verifier::new_with_fake_root(ROOT1); // TODO: time
@@ -281,6 +306,10 @@ fn test_with_mock_root(test_case: &TestCase) {
         std::time::SystemTime::now(),
     );
 
-    assert_cert_error_eq(&result.map(|_| ()), &test_case.expected_result);
+    assert_cert_error_eq(
+        &result.map(|_| ()),
+        &test_case.expected_result,
+        test_case.other_error.as_ref(),
+    );
     // TODO: get into specifics of errors returned when it fails.
 }
diff --git a/src/tests/verification_real_world/mod.rs b/src/tests/verification_real_world/mod.rs
index 12e0e92..4d1d26e 100644
--- a/src/tests/verification_real_world/mod.rs
+++ b/src/tests/verification_real_world/mod.rs
@@ -113,7 +113,13 @@ macro_rules! real_world_test_cases {
     }
 }
 
-fn real_world_test(test_case: &TestCase) {
+macro_rules! no_error {
+    () => {
+        None::<std::convert::Infallible>
+    };
+}
+
+fn real_world_test<E: std::error::Error>(test_case: &TestCase<E>) {
     log::info!("verifying {:?}", test_case.expected_result);
 
     let verifier = crate::verifier_for_testing();
@@ -141,7 +147,11 @@ fn real_world_test(test_case: &TestCase) {
         )
         .map(|_| ());
 
-    assert_cert_error_eq(&result.map(|_| ()), &test_case.expected_result);
+    assert_cert_error_eq(
+        &result.map(|_| ()),
+        &test_case.expected_result,
+        None::<&std::convert::Infallible>,
+    );
     // TODO: get into specifics of errors returned when it fails.
 }
 
@@ -154,6 +164,7 @@ real_world_test_cases! {
         chain: VALID_1PASSWORD_COM_CHAIN,
         stapled_ocsp: None,
         expected_result: Ok(()),
+        other_error: no_error!(),
     },
     // Same as above but without stapled OCSP.
     my_1password_com_valid_no_stapled => TestCase {
@@ -161,6 +172,7 @@ real_world_test_cases! {
         chain: VALID_1PASSWORD_COM_CHAIN,
         stapled_ocsp: None,
         expected_result: Ok(()),
+        other_error: no_error!(),
     },
     // Valid also for 1password.com (no subdomain).
     _1password_com_valid => TestCase {
@@ -168,6 +180,7 @@ real_world_test_cases! {
         chain: VALID_1PASSWORD_COM_CHAIN,
         stapled_ocsp: None,
         expected_result: Ok(()),
+        other_error: no_error!(),
     },
     // The certificate isn't valid for an unrelated subdomain.
     unrelated_domain_invalid => TestCase {
@@ -175,6 +188,7 @@ real_world_test_cases! {
         chain: VALID_1PASSWORD_COM_CHAIN,
         stapled_ocsp: None,
         expected_result: Err(TlsError::InvalidCertificate(CertificateError::NotValidForName)),
+        other_error: no_error!(),
     },
     // The certificate chain for the unrelated domain is valid for that
     // unrelated domain.
@@ -183,6 +197,7 @@ real_world_test_cases! {
         chain: VALID_UNRELATED_CHAIN,
         stapled_ocsp: None,
         expected_result: Ok(()),
+        other_error: no_error!(),
     },
     // The certificate chain for the unrelated domain is not valid for
     // my.1password.com.
@@ -191,6 +206,7 @@ real_world_test_cases! {
         chain: VALID_UNRELATED_CHAIN,
         stapled_ocsp: None,
         expected_result: Err(TlsError::InvalidCertificate(CertificateError::NotValidForName)),
+       other_error: no_error!(),
     },
 
     // OCSP stapling works.
diff --git a/src/verification/android.rs b/src/verification/android.rs
index e68ee38..9bd0ff1 100644
--- a/src/verification/android.rs
+++ b/src/verification/android.rs
@@ -9,7 +9,7 @@ use std::time::SystemTime;
 
 use super::{log_server_cert, unsupported_server_name, ALLOWED_EKUS};
 use crate::android::{with_context, CachedClass};
-use crate::verification::{error_messages, invalid_certificate};
+use crate::verification::invalid_certificate;
 
 static CERT_VERIFIER_CLASS: CachedClass =
     CachedClass::new("com/onepassword/rustls_platform_verifier/CertificateVerifier");
@@ -216,9 +216,9 @@ impl Verifier {
                     VerifierStatus::InvalidEncoding => {
                         Err(InvalidCertificate(CertificateError::BadEncoding))
                     }
-                    VerifierStatus::InvalidExtension => {
-                        Err(invalid_certificate(error_messages::INVALID_EXTENSIONS))
-                    }
+                    VerifierStatus::InvalidExtension => Err(InvalidCertificate(
+                        CertificateError::Other(std::sync::Arc::new(super::EkuError)),
+                    )),
                 }
             }
             Err(e) => Err(TlsError::General(format!(
diff --git a/src/verification/apple.rs b/src/verification/apple.rs
index 5bf3b25..02edb68 100644
--- a/src/verification/apple.rs
+++ b/src/verification/apple.rs
@@ -1,5 +1,5 @@
 use super::{log_server_cert, unsupported_server_name};
-use crate::verification::{error_messages, invalid_certificate};
+use crate::verification::invalid_certificate;
 use core_foundation::date::CFDate;
 use core_foundation_sys::date::kCFAbsoluteTimeIntervalSince1970;
 use rustls::{client::ServerCertVerifier, CertificateError, Error as TlsError};
@@ -164,9 +164,9 @@ impl Verifier {
                     errors::errSecCreateChainFailed => Ok(TlsError::InvalidCertificate(
                         CertificateError::UnknownIssuer,
                     )),
-                    errors::errSecInvalidExtendedKeyUsage => {
-                        Ok(invalid_certificate(error_messages::INVALID_EXTENSIONS))
-                    }
+                    errors::errSecInvalidExtendedKeyUsage => Ok(TlsError::InvalidCertificate(
+                        CertificateError::Other(std::sync::Arc::new(super::EkuError)),
+                    )),
                     errors::errSecCertificateRevoked => {
                         Ok(TlsError::InvalidCertificate(CertificateError::Revoked))
                     }
diff --git a/src/verification/mod.rs b/src/verification/mod.rs
index f20e7e0..de51e85 100644
--- a/src/verification/mod.rs
+++ b/src/verification/mod.rs
@@ -22,11 +22,18 @@ mod windows;
 #[cfg(windows)]
 pub use windows::Verifier;
 
-#[cfg(any(windows, target_os = "android", target_os = "macos", target_os = "ios"))]
-/// Custom error messages for errors that are the same across non-WebPKI verification platforms.
-pub mod error_messages {
-    pub const INVALID_EXTENSIONS: &str = "certificate had invalid extensions";
+/// An EKU was invalid for the use case of verifying a server certificate.
+///
+/// This error is used primarily for tests.
+#[derive(Debug, PartialEq)]
+pub(crate) struct EkuError;
+
+impl std::fmt::Display for EkuError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.write_str("certificate had invalid extensions")
+    }
 }
+impl std::error::Error for EkuError {}
 
 // Log the certificate we are verifying so that we can try and find what may be wrong with it
 // if we need to debug a user's situation.
diff --git a/src/verification/others.rs b/src/verification/others.rs
index 3505eae..f7f9d27 100644
--- a/src/verification/others.rs
+++ b/src/verification/others.rs
@@ -2,7 +2,7 @@ use super::log_server_cert;
 use once_cell::sync::OnceCell;
 use rustls::{
     client::{ServerCertVerifier, WebPkiVerifier},
-    Error as TlsError, RootCertStore,
+    CertificateError, Error as TlsError, RootCertStore,
 };
 
 #[derive(Default)]
@@ -119,6 +119,7 @@ impl ServerCertVerifier for Verifier {
                 ocsp_response,
                 now,
             )
+            .map_err(map_webpki_errors)
             // This only contains information from the system or other public
             // bits of the TLS handshake, so it can't leak anything.
             .map_err(|e| {
@@ -128,6 +129,20 @@ impl ServerCertVerifier for Verifier {
     }
 }
 
+fn map_webpki_errors(err: TlsError) -> TlsError {
+    if let TlsError::InvalidCertificate(CertificateError::Other(other_err)) = &err {
+        if let Some(webpki_error) = other_err.downcast_ref::<webpki::Error>() {
+            if let webpki::Error::RequiredEkuNotFound = webpki_error {
+                return TlsError::InvalidCertificate(CertificateError::Other(std::sync::Arc::new(
+                    super::EkuError,
+                )));
+            }
+        }
+    }
+
+    err
+}
+
 /// Loads the static `webpki-roots` into the provided certificate store.
 fn load_webpki_roots(store: &mut RootCertStore) {
     use rustls::OwnedTrustAnchor;
diff --git a/src/verification/windows.rs b/src/verification/windows.rs
index 8e2167d..7ae5f24 100644
--- a/src/verification/windows.rs
+++ b/src/verification/windows.rs
@@ -52,7 +52,7 @@ use std::{
     time::SystemTime,
 };
 
-use crate::verification::{error_messages, invalid_certificate};
+use crate::verification::invalid_certificate;
 #[cfg(any(test, feature = "ffi-testing", feature = "dbg"))]
 use winapi::um::wincrypt::CERT_CHAIN_ENGINE_CONFIG;
 
@@ -455,7 +455,9 @@ fn map_trust_error_status(unfiltered_status: DWORD) -> Result<(), TlsError> {
         wincrypt::CERT_TRUST_IS_NOT_VALID_FOR_USAGE
             | wincrypt::CERT_TRUST_CTL_IS_NOT_VALID_FOR_USAGE,
     ) {
-        return Err(invalid_certificate(error_messages::INVALID_EXTENSIONS));
+        return Err(InvalidCertificate::Other(std::sync::Arc::new(
+            super::EkuError,
+        )));
     }
 
     // Otherwise, if there is only one class of error, map that class to

```

cpu: Thanks for the patch! I will apply this and your other feedback ASAP.

src/tests/verification_mock/mod.rs
cpu: Fixed!

src/verification/windows.rs
cpu: Fixed!

src/tests/mod.rs
cpu: This worked nicely :+1: 

I landed this in-branch with a couple minor tweaks (a clippy fix related to nested `if let`, and turning off a dead code warning on the `other_error` field for cfgs that weren't using it).

cpu:
@complexspaces I think this is ready for another review pass when you have a chance. There's no big rush because we still have a few other dependencies that need to be updated so we can remove the Cargo patches here.

complexspaces (approved):
I have one last thing, but otherwise this looks good. Now its just a matter of the rest of the ecosystem catching up :)

src/tests/verification_mock/mod.rs
```diff
@@ -177,60 +194,80 @@ mock_root_test_cases! {
...
+            CertificateError::Other(Arc::from(EkuError)))),
+        other_error: Some(EkuError),
+    },
+    wrong_eku_dns [ any(windows, target_os = "android", target_os = "macos") ] => TestCase {
+        reference_id: EXAMPLE_COM,
+        chain: &[include_bytes!("root1-int1-ee_example.com-wrong_eku.crt"), ROOT1_INT1],
+        stapled_ocsp: None,
+        expected_result: Err(TlsError::InvalidCertificate(
+            CertificateError::Other(Arc::from(EkuError)))),
+        other_error: Some(EkuError),
```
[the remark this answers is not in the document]
cpu: Good catch! I collapsed these back into one test case in my last force push.

src/tests/mod.rs
```diff
@@ -15,5 +19,30 @@ struct TestCase<'a> {
...
     pub stapled_ocsp: Option<&'a [u8]>,
 
-    pub expected_result: Result<(), rustls::Error>,
+    pub expected_result: Result<(), TlsError>,
+
+    /// An error that should be present inside an expected `CertificateError::Other` variant.
+    ///
+    /// Set this if the error being tested uses `CertificateError::Other` and not statically known
+    /// variants in [TlsError]
+    #[allow(dead_code)]
```
cpu, line 28: on a lark I tried to replace this with a `cfg` directive that excluded the field from the MacOS builds that weren't using it. That ends up just shifting the problem as the `E` generic parameter isn't used at that point and we'd need a `PhantomData` field or some other reworking.

TLDR: It felt to me like `allow(dead_code)` was the right tool for this particular instance.

complexspaces: Yup, I agree 👍. Especially in test code I prefer to just silence dead code warnings if the alternative ends up being significantly more complex. [Here](https://github.com/rustls/rustls-platform-verifier/blob/main/src/tests/ffi.rs#L132) is an example of that, where figuring out exact combinations of test-only feature flags wasn't worth the noise.

cpu:
> cpu force-pushed the cpu-rustls-0.21.0-prep branch from cf59e4b to dda43b4 

Rebased on main and fixed the Cargo.toml conflict.

> Now its just a matter of the rest of the ecosystem catching up :)

Slowly but surely :-) I updated the README with a rough checklist of things we're waiting on.

complexspaces:
Thanks for the list, that's pretty helpful 👍

cpu:
Haven't forgotten about this PR :-) Things are moving again and I'm hopeful we'll be able to tie a bow on this in the next week or two.

complexspaces:
No worries at all, I've been following along with the other releases that are leading up to this.

cpu:
@complexspaces All set! :rocket:

Cargo.toml
```diff
@@ -47,10 +47,11 @@ once_cell = { version = "1.9", optional = true } # Only used during doc generati
 rustls-native-certs = "0.6"
 once_cell = "1.9"
 webpki-roots = "0.22" # Augmentation to `openssl-probe` due to inconsistencies on Linux distros.
```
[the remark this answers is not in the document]
cpu: Good catch, thanks. Updated.

complexspaces:
Thanks again for handling this!

PR #10, opened 2023-03-27
Import 1Password downstream changes

complexspaces:
This PR imports all the changes that we (1Password) had applied to our local version of this crate in order to fix issues we encountered in production.

These include:
- Fixing a crash while verifying a certificate on macOS. We're not sure why this was actually happening and we're unable to narrow the cause down, but we suspect someone's system clock or network setup was introducing a bad timestamp somewhere. A [short discussion](https://discord.com/channels/976380008299917365/1015156984007381033/1074853630613659689) on the rustls Discord server agreed that "...probably makes sense to error out" regardless.
- Unconditionally including the trust anchors from `webpki-roots` on Linux for certificate verification. These are now always mixed into the certificates loaded from the system (if any). In production we've found `openssl-probe` to be fairly unreliable on older installations of even standard distros, leading to no roots for our server's (one of Amazon's Root CAs) certificate being found.

I believe that only the second change would be controversial, but AFAIK we still need to rework how `webpki-roots` is used at all in this crate so I don't feel its that big of a deal at the moment and would help accelerate those.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -46,7 +46,7 @@ once_cell = { version = "1.9", optional = true } # Only used during doc generati
 [target.'cfg(target_os = "linux")'.dependencies]
 rustls-native-certs = "0.6"
 once_cell = "1.9"
-webpki-roots = "0.22" # Fallback when `openssl-probe` can't find anything.
+webpki-roots = "0.22" # Augmentation to `openssl-probe` due to inconsistencies on Linux distros.
 
 [target.'cfg(target_os = "android")'.dependencies]
 jni = { version = "0.19", default-features = false }
--- a/src/verification/apple.rs
+++ b/src/verification/apple.rs
@@ -27,15 +27,13 @@ fn system_time_to_cfdate(time: SystemTime) -> Result<CFDate, TlsError> {
 
     // Convert a system timestamp based off the UNIX epoch into the
     // Apple epoch used by all `CFAbsoluteTime` values.
-
-    let since_unix_epoch = time
-        .duration_since(SystemTime::UNIX_EPOCH)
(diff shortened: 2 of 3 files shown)


cpu:
> These are now always mixed into the certificates loaded from the system (if any).

Is this always an additive process? Could a native root from the system root store be replaced by something from webpki-roots?

> I believe that only the second change would be controversial, but AFAIK we still need to rework how webpki-roots is used at all in this crate so I don't feel its that big of a deal at the moment and would help accelerate those.

Can you expand on the need to rework how webpki-roots is being used? I probably missed any discussions about that.

This feels like a reasonable set of changes but I want to make sure I'm not missing broader context.

complexspaces:
> Is this always an additive process? Could a native root from the system root store be replaced by something from webpki-roots?

Yes, its always additive. Trust anchors are just appended to a list without any deduplication. Even if they had more handling applied, one would hope that a root with a certain SPKI would be the same in both places 😓

> Can you expand on the need to rework how webpki-roots is being used? I probably missed any discussions about that.

Sure, that's my bad for not elaborating more. The previous discussions happened when the crate first got put up here on GitHub and revolved around if we even want to have `webpki-roots` as a dependency of `rustls-platform-verifier`, or in what fashion it should remain. I believe this is the final approach we settled on, to balance convenience with misuse resistance:
> The use of webpki-roots can be an optional feature enabled with explicit API

After writing all that out, I should really get this into its own issue for tracking.

cpu (approved):
These seem like reasonable changes to me. Thanks @complexspaces !

complexspaces:
I made a small change to keep around the error log about the system not providing any CA roots. I think its helpful to have for troubleshooting, even if a connection still works.

PR #11, opened 2023-03-27
Redo android dependency management

complexspaces:
This PR overhauls how we distribute and compile the Kotlin component of the library, which is required to run on Android. It makes it significantly less complicated and less fragile. These changes remove the previous strategy of trying to "install" the Gradle module into a consumer's project working directory and instead simply locate the copy stored inside cargo's downloaded crate cache.

To provide some background context on why even we need this: The "native" Android certificate verification APIs are implemented in  [conscrypt](https://github.com/google/conscrypt), which is a library that provides an implementation of Java's certificate interfaces/classes/etc. Since the library is private and not exposed to third-party applications, the "correct" way to verify certificates natively on the platform is to use the Java APIs. There's two options to do that from Rust:
- Pure JNI code, written 100% in Rust.
- Rust JNI wrappers with a Java implementation compiled into the application.

The first one is extremely verbose and hard to maintain, so we went with the second (also mirroring what Chromium and other large projects do). So in order for everything to work correctly, we need to make sure the consumer app is compiling our Kotlin source into their build so that the class and methods will be available to call via JNI.

The prior approach I took was bad for a few reason:
- Writing to a random place in the target directory is generally discouraged by `cargo`.
- A build script was required. On top of that, by design, it wasn't able to correctly track all its inputs in fear of triggering spurious Rust rebuilds.
- Writing a whole Gradle project/module to the Rust target directory (or even the Gradle generated source directory) caused issues with Android Studio's rebuild functionality and was generally flaky because its hard to make sure the project definition is always there when Gradle goes to load it.

This new implementation should require very little to no maintenance once setup and is free of all the aforementioned issues. We've also tested it internally by integrating it into 1Password 8 for Android's build system to good success.


complexspaces:
@cpu Did you want to look at this one as well before I merge it? Its primarily Gradle changes which I'm unsure of your familiarity with.

cpu:
> Its primarily Gradle changes which I'm unsure of your familiarity with.

@complexspaces Sorry, zero familiarity with Gradle :sweat: I probably won't have a chance to take a look at this diff immediately since it will take a lot of background reading to be able to give you any useful feedback. If you're really keen on a second set of eyes I can try and make time ~Friday.

complexspaces:
I wouldn't say this one is super critical to review if its not already something your familiar with, especially with the time it sounds like you'd need to sink in for a single MR. I don't want to take away from your other tasks just for this. One of the Android developers at 1P also had a look at this approach and didn't have any issues.

PR #19, opened 2023-07-31
Use valid Java package names

### LINKED ISSUE #18: Invalid Java package name
Currently, the Android library uses a package name of `com.onepassword.rustls-platform-verifier`, which works fine on older versions of Java but on newer versions of Java presents us with this build failure:

> Package 'rustls.rustls-platform-verifier' from AndroidManifest.xml is not a valid Java package name as 'rustls-platform-verifier' is not a valid Java identifier.

Underscores are not considered valid characters in a Java identifier.

**Suggestion:** change the package name to something like `rustls.platformverifier.android` (not sure why 1Password needs to be referenced at all) and remove all of the underscores.

This is currently blocking some Android updates at 1Password so I'm happy to make this change but I just wanted to file an issue so that this is being tracked.

jonduran3000:
Updated the Android code to use valid Java packages names. The package names aren't a problem when using older versions of Java but newer versions of Java validate the names and cause build failures.

Resolves #18


complexspaces:
I will take a look at CI soon and figure out why its failing. The failures I am seeing do not look related to your changes.

complexspaces:
Once https://github.com/rustls/rustls-platform-verifier/pull/20 merges and fixes CI, this should be good to merge.

complexspaces (approved):

PR #20, opened 2023-08-01
Update test certificates

complexspaces:
This PR updates both the real-world certificates used for testing and then mock CAs and their leaf certificates.


cpu (approved):
I didn't decode each of the updated test files to inspect the contents but instead am relying on the unit tests continuing to pass as a demonstration that this branch's diff is sensible. I hope that's reasonable :-)

complexspaces:
Thanks for the review, and that also seems reasonable to me. I didn't change any actual test logic so bogus certificates would fail. In this case, for reference, I used the existing script and Firefox's certificate viewer to grab the new ones.

Issue #22, opened 2023-08-25
jni 0.21 migration


cpu:
Presently this crate uses [`jni-rs`](https://github.com/jni-rs/jni-rs) v0.19 as an optional dependency for doc generation, and a required dependency for the Android OS target.

The latest release stream, 0.21.x, "makes extensive breaking changes in order to improve safety. Most projects that use this library will need to be changed accordingly.". The upstream project offers a [migration guide](https://github.com/jni-rs/jni-rs/blob/master/docs/0.21-MIGRATION.md) to help with this process.

Given the advertised safety improvements it's likely worthwhile to consider the steps required for migration.

complexspaces:
Thanks for taking a glance at this! The new safety improvements look very welcome. Our uses are pretty simple so I'm hoping our upgrade path is pretty smooth.

djc:
Yes.

PR #23, opened 2023-08-25
misc. dependency updates

cpu:
Before diving into more meaningful work I wanted to update some dependencies (_possibly a good time to consider adding a Dependabot config?_). This branch updates all of the deps and dev deps (with the exception of `jni` (see https://github.com/rustls/rustls-platform-verifier/issues/22), and semver compatible dependencies). Likely easiest to review commit-by-commit.

Additionally rolls in a small `.gitignore` tweak for CLion users and a tweak to the CI config to make it easier to run CI without opening a PR or making config changes.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -47,11 +47,11 @@ once_cell = { version = "1.9", optional = true } # Only used during doc generati
 rustls-native-certs = "0.6"
 once_cell = "1.9"
 webpki-roots = "0.25" # Augmentation to `openssl-probe` due to inconsistencies on Linux distros.
-webpki = { package = "rustls-webpki", version = "0.100.0", features = ["alloc", "std"] }
+webpki = { package = "rustls-webpki", version = "0.101", features = ["alloc", "std"] }
 
 [target.'cfg(target_os = "android")'.dependencies]
 jni = { version = "0.19", default-features = false }
-webpki = { package = "rustls-webpki", version = "0.100.0", features = ["alloc", "std"] }
+webpki = { package = "rustls-webpki", version = "0.101", features = ["alloc", "std"] }
 once_cell = "1.9"
 android_logger = { version = "0.11", optional = true } # Only used during testing.
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -53,7 +53,7 @@ webpki = { package = "rustls-webpki", version = "0.101", features = ["alloc", "s
 jni = { version = "0.19", default-features = false }
 webpki = { package = "rustls-webpki", version = "0.101", features = ["alloc", "std"] }
 once_cell = "1.9"
-android_logger = { version = "0.11", optional = true } # Only used during testing.
+android_logger = { version = "0.13", optional = true } # Only used during testing.
 
 [target.'cfg(target_arch = "wasm32")'.dependencies]
 once_cell = "1.9"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -39,7 +39,7 @@ docsrs = ["jni", "once_cell"]
 [dependencies]
 rustls = { version = "0.21", features = ["dangerous_configuration", "tls12", "logging"] }
 log = { version = "0.4" }
-base64 = { version = "0.13", optional = true } # Only used when the `cert-logging` feature is enabled.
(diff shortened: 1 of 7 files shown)


Cargo.toml
```diff
@@ -39,38 +39,38 @@ docsrs = ["jni", "once_cell"]
 [dependencies]
 rustls = { version = "0.21", features = ["dangerous_configuration", "tls12", "logging"] }
 log = { version = "0.4" }
-base64 = { version = "0.13", optional = true } # Only used when the `cert-logging` feature is enabled.
+base64 = { version = "0.21", optional = true } # Only used when the `cert-logging` feature is enabled.
 jni = { version = "0.19", default-features = false, optional = true } # Only used during doc generation
-once_cell = { version = "1.9", optional = true } # Only used during doc generation.
+once_cell = { version = "1.18", optional = true } # Only used during doc generation.
```
djc, line 44: 1.9 and 1.18 are semver-compatible, so I see no reason to shrink the compatibility range?

Cargo.toml
```diff
@@ -39,38 +39,38 @@ docsrs = ["jni", "once_cell"]
...
+security-framework = { version = "2.9", features = ["OSX_10_14"] }
+security-framework-sys = { version = "2.9", features = ["OSX_10_14"] }
 
 [target.'cfg(windows)'.dependencies]
 winapi = { version = "0.3", features = ["wincrypt", "winerror"] }
 
 [dev-dependencies]
-tokio = { version = "1.5", features = ["macros", "rt-multi-thread"] }
-reqwest = { version = "0.11.18", default-features = false, features = ["rustls-tls-manual-roots"] }
+tokio = { version = "1.32", features = ["macros", "rt-multi-thread"] }
```
djc, line 72: These are semver-compatible.

Cargo.toml
```diff
@@ -39,38 +39,38 @@ docsrs = ["jni", "once_cell"]
...
-webpki-roots = "0.23"
+once_cell = "1.18"
+webpki-roots = "0.25"
 
 [target.'cfg(any(target_os = "macos", target_os = "ios"))'.dependencies]
 core-foundation = "0.9"
 core-foundation-sys = "0.8"
-security-framework = { version = "2.6", features = ["OSX_10_14"] }
-security-framework-sys = { version = "2.4", features = ["OSX_10_14"] }
+security-framework = { version = "2.9", features = ["OSX_10_14"] }
```
djc, line 65: Semver-compatible.

djc (approved):
Increasing the requirements within a semver-compatible range is unnecessary, so I'd avoid most of these bumps.

cpu:
> Increasing the requirements within a semver-compatible range is unnecessary, so I'd avoid most of these bumps.

Good point :sweat_smile: I'll remove those.

complexspaces:
I agree with @djc on some of these bumps, especially for `tokio`. The CI changes look fine though 👍.

While it shouldn't be the sole decision making process, its become increasingly hard at work to make some of these increases without mass dependency duplication. So, leaving the ranges wider makes it easier to match ones that have more compatible sub-dependencies. For example this seems to duplicate `socket2` and `windows-sys` and would push those downstream as well.

Cargo.toml
cpu: Removed.

Cargo.toml
cpu: Removed.

Cargo.toml
cpu: Removed.

cpu:
Fixed up to remove the semver compatible bumps.

complexspaces:
@cpu Do you mind also running `cargo update -p rustls` before merging to drop the duplicate `rustls-webpki` dependency from the lockfile? Otherwise CI will need to build it twice every time.

cpu:
@complexspaces Ahhh! Yes, done :ballot_box_with_check:  

I was prev. confused why after re-arranging some of the commits I had to remove the deprecated `add_server_trust_anchors` fix I had landed in the first go, but it was related to this duplication and the lockfile `rustls` version. 

All sorted now, thank you.

complexspaces (approved):
I sanity checked these changes against another codebase and confirmed nothing blew up. LGTM 😄

PR #24, opened 2023-08-25
Expose Verifier in API, remove Linux webpki-roots dependency.

### LINKED ISSUE #7: expose the Verifier struct
Hi, thanks for the great work.

I am wondering if the Verifier struct be made public, I want to configure my own `ClientConfig`, for example, customize preferred cipher suites.

### LINKED ISSUE #12: Make Linux's webpki-roots optional and opt-in at runtime
Currently, the use of `webpki-roots` in the [Linux verifier](https://github.com/rustls/rustls-platform-verifier/blob/main/src/verification/others.rs#L70-L78) is not configurable and they are always loaded to supplement what might come from on-disk CA bundles. This, and the previous behavior of falling back to `webpki-roots`, was implemented at 1Password because we found `openssl-probe` to be way too unreliable and had multiple users write in that they were unable to use the app as a result.

However, this is likely not what everyone wants and makes it harder to control exactly what trust roots a consumer of `rustls-platform-verifier` might allow in addition to some other concerns:

> I think webpki-roots should not be included on Linux, since they will inevitably bitrot - @Ralith

But there we're also some solutions presented as well:

> I have mixed feelings about a feature, because those can end up silently enabled, which is a bit of a security hazard
maybe support manually adding certs and let people pipe in webpki-roots themselves if they really want to? - @Ralith

> The use of webpki-roots can be an optional feature enabled with explicit API - @djc

Out of these, the last idea to enable it with an explicit API seems to be the best balance between convenience, a simple crate API, and control/security. So I believe that is the way we should go.

There are then two things needed to resolve this:
- Make all `webpki-roots` dependencies optional via an opt-in feature called `webpki-roots` or `webpki-roots-augment`.
- Add an API to the top-level crate like this which sets a global atomic and then update `others.rs` to use it on Linux:
```rust
pub fn enable_webpki_fall() {
   ...
}

### PR DESCRIPTION
**Pull request by cpu**
## Description

This branch builds on https://github.com/rustls/rustls-platform-verifier/pull/23 and so should be reviewed commit-by-commit starting at b4db665fec0f24382bbfb9be6bbe6053dc77bf0c - the commits prior to this SHA are from #23.

The set of changes in this branch aim to accomplish two related goals:

1. Making the per-platform `Verifier` struct public, resolving https://github.com/rustls/rustls-platform-verifier/issues/7
2. Removing the Linux platform dependency on `webpki-roots`, replacing it with a new `Verifier::new_with_extra_roots` constructor, resolving https://github.com/rustls/rustls-platform-verifier/issues/12.
Callers can choose to use this constructor to provide roots from `webpki-roots` (or another source) and they will be added to the constructed `RootCertStore` in addition to trust anchors loaded from `rustls-native-certs`.

The `wasm32` target continues to depend on `webpki-roots` since there is no `rustls-native-certs` support for that target.

Resolves #7.
Resolves #12.

### TODO:

* [x] Land #23
* [x] Rebase

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -46,7 +46,6 @@ once_cell = { version = "1.9", optional = true } # Only used during doc generati
 [target.'cfg(target_os = "linux")'.dependencies]
 rustls-native-certs = "0.6"
 once_cell = "1.9"
-webpki-roots = "0.25" # Augmentation to `openssl-probe` due to inconsistencies on Linux distros.
 webpki = { package = "rustls-webpki", version = "0.101", features = ["alloc", "std"] }
 
 [target.'cfg(target_os = "android")'.dependencies]
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -6,7 +6,7 @@ use rustls::ClientConfig;
 use std::sync::Arc;
 
 mod verification;
-use verification::Verifier;
+pub use verification::Verifier;
 
 // Build the Android module when generating docs so that
 // the Android-specific functions are included.
@@ -28,7 +28,8 @@ mod tests;
 pub use tests::ffi::*;
 
 /// Creates and returns a `rustls` configuration that verifies TLS
-/// certificates in the best way for the underlying OS platform.
+/// certificates in the best way for the underlying OS platform, using
+/// safe defaults for the `rustls` configuration.
 ///
 /// # Example
 ///
@@ -45,6 +46,10 @@ pub use tests::ffi::*;
 ///     let _response = client.get("https://example.com").send().await;
 /// }
 /// ```
+///
+/// If you require more control over the rustls `ClientConfig`, you can
+/// instantiate a [Verifier] with [Verifier::default] and then use it
+/// with [rustls::ConfigBuilder::with_custom_certificate_verifier].
 ///
 /// Refer to the crate level documentation to see what platforms
 /// are currently supported.
--- a/src/verification/android.rs
+++ b/src/verification/android.rs
@@ -35,6 +35,7 @@ enum VerifierStatus {
 // official recommendation. See https://bugs.chromium.org/p/chromium/issues/detail?id=627154.
 const AUTH_TYPE: &str = "RSA";
 
+#[derive(Default)]
 pub struct Verifier {
     /// Testing only: The root CA certificate to trust.
     #[cfg(any(test, feature = "ffi-testing"))]
--- a/src/verification/android.rs
+++ b/src/verification/android.rs
@@ -64,7 +64,7 @@ impl Verifier {
     }
 
     #[cfg(any(test, feature = "ffi-testing"))]
-    pub fn new_with_fake_root(root: &[u8]) -> Self {
+    pub(crate) fn new_with_fake_root(root: &[u8]) -> Self {
         Self {
             test_only_root_ca_override: Some(root.into()),
         }
--- a/src/verification/android.rs
+++ b/src/verification/android.rs
@@ -70,7 +70,7 @@ impl Verifier {
         }
     }
 
-    pub fn verify_certificate(
+    fn verify_certificate(
         &self,
         end_entity: &rustls::Certificate,
         intermediates: &[rustls::Certificate],
--- a/src/verification/android.rs
+++ b/src/verification/android.rs
@@ -35,6 +35,7 @@ enum VerifierStatus {
 // official recommendation. See https://bugs.chromium.org/p/chromium/issues/detail?id=627154.
 const AUTH_TYPE: &str = "RSA";
 
+/// A TLS certificate verifier that utilizes the Android platform verifier.
 #[derive(Default)]
 pub struct Verifier {
     /// Testing only: The root CA certificate to trust.
@@ -56,6 +57,8 @@ impl Drop for Verifier {
 }
 
 impl Verifier {
+    /// Creates a new instance of a TLS certificate verifier that utilizes the
+    /// Android certificate facilities
     pub fn new() -> Self {
         Self {
             #[cfg(any(test, feature = "ffi-testing"))]
@@ -63,6 +66,7 @@ impl Verifier {
         }
     }
 
+    /// Creates a test-only TLS certificate verifier which trusts our fake root CA cert.
     #[cfg(any(test, feature = "ffi-testing"))]
     pub(crate) fn new_with_fake_root(root: &[u8]) -> Self {
         Self {
--- a/src/verification/apple.rs
+++ b/src/verification/apple.rs
@@ -37,6 +37,7 @@ fn system_time_to_cfdate(time: SystemTime) -> Result<CFDate, TlsError> {
         .map(|epoch| CFDate::new(epoch.as_secs() as f64))
 }
 
+/// A TLS certificate verifier that utilizes the Apple platform certificate facilities.
 #[derive(Default)]
 pub struct Verifier {
     /// Testing only: The root CA certificate to trust.
--- a/src/verification/others.rs
+++ b/src/verification/others.rs
@@ -2,8 +2,9 @@ use super::log_server_cert;
 use once_cell::sync::OnceCell;
 use rustls::{
     client::{ServerCertVerifier, WebPkiVerifier},
-    CertificateError, Error as TlsError, RootCertStore,
+    CertificateError, Error as TlsError,
 };
+use std::sync::Mutex;
 
 /// A TLS certificate verifier that uses the system's root store and WebPKI.
 #[derive(Default)]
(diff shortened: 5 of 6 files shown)

### CODE REVIEW

**Comment by cpu** [comment]
> cpu marked this pull request as draft now

Doing this for two reasons:
1. To focus review on #23 first.
2. To confirm this approach matches what @complexspaces et. al. were thinking in #12.

**Comment by djc** [comment]
For stacked branches, consider setting the target branch to the parent. That way, you don't see the extra commits and GitHub will change the target after the parent branch is merged.

File: Cargo.toml [hunk 1]
```diff
@@ -35,6 +35,10 @@ dbg = []
 cert-logging = ["base64"]
 # Used for nicely documenting the Android-specific APIs. This feature is not stable.
 docsrs = ["jni", "once_cell"]
+# Enables API for Linux verifier to be provided extra non-platform-provided roots (e.g.
+# from `webpki-roots`). This can be helpful to augment the platform-provided roots
+# when the platform-provided roots are out of date.
+extra-roots = []
```
Ralith, line 41: This seems like a small, portable API. Why bother masking it behind a feature?

cpu:
> For stacked branches, consider setting the target branch to the parent. That way, you don't see the extra commits and GitHub will change the target after the parent branch is merged.

I think that only works if you push the parent branch to the origin repo, otherwise it will open a PR on my fork and not ping any reviewers. I'm not generally in the habit of pushing branches to the primary repo (they can end up collecting into messy clutter, and it's only possible with maintainer access) but I could consider it for these stacked changes.

Ralith (approved):
Seems like a great solution!

Cargo.toml
Ralith: I understand djc to be thinking of a webpki-specific API, but I think the less opinionated approach you have here is preferable in part *because* there's no need to gate it.

Ralith:
Alternatively, rustls(?) could provide a wrapper for any `ServerCertVerifier` that adds additional trust roots, by checking against them before delegating to the inner verifier. That'd be a bit more general in principle, though less discoverable and perhaps not widely useful.

src/verification/android.rs
```diff
@@ -56,13 +57,16 @@ impl Drop for Verifier {
...
     pub fn new() -> Self {
         Self {
             #[cfg(any(test, feature = "ffi-testing"))]
             test_only_root_ca_override: None,
         }
     }
 
+    /// Creates a test-only TLS certificate verifier which trusts our fake root CA cert.
     #[cfg(any(test, feature = "ffi-testing"))]
     pub fn new_with_fake_root(root: &[u8]) -> Self {
```
complexspaces, line 71: Suggestion: I believe this can be made `pub(crate)` like the other platforms. This is a leftover from when the tests which used this were outside of `src`.

src/verification/apple.rs
```diff
@@ -37,6 +37,7 @@ fn system_time_to_cfdate(time: SystemTime) -> Result<CFDate, TlsError> {
         .map(|epoch| CFDate::new(epoch.as_secs() as f64))
 }
 
+/// A TLS certificate verifier that uses the system's root store and WebPKI.
```
complexspaces, line 40: I don't believe this is correct, was this a copy/paste error from the `other` verifier? We use both the system's root store and Apple's TLS verification routines that come from `Security.framework`. This one should be a closer match to the comment for Windows.

src/lib.rs
```diff
@@ -46,6 +46,10 @@ pub use tests::ffi::*;
 /// }
 /// ```
 ///
+/// If you require more control over the rustls `ClientConfig`, you can
```
complexspaces, line 49: Nit: Maybe leave an extra comment at the top of the documentation block that mentions it uses safe defaults so people have an example of why they might need tighter control.

Cargo.toml
complexspaces: I like the idea of always exposing it. That would also match up better with [the API in `rustls` itself](https://docs.rs/rustls/latest/rustls/struct.ConfigBuilder.html#method.with_root_certificates) which is always available even without the dangerous configuration feature enabled.

djc:
I was thinking in terms of a specific API call that makes it trivial to add the webpki-roots specifically, but maybe the we can make that easy enough anyway by way of the pki-types project?

Cargo.toml
djc: I generally wouldn't mask extra API behind a feature unless it requires extra dependencies or there is another motivation (like for dangerous configuration). I agree with @Ralith that it probably doesn't make sense here.

src/verification/others.rs
```diff
@@ -68,35 +89,40 @@ impl Verifier {
...
-                // addition so that a valid trust anchor is more likely to be available.
-                load_webpki_roots(&mut root_store);
-
-                log::debug!(
-                    "Loaded WebPKI roots in addition to {} roots from the system",
-                    added
-                );
+                #[cfg(feature = "extra-roots")]
+                {
+                    root_store.add_server_trust_anchors(&mut self.extra_roots.clone().into_iter());
```
complexspaces, line 100: Nit: It would be nice to not need to duplicate these roots twice in memory for the lifetime of the verifier (three times if they're also hardcoded in the binary like `webpki-roots`. I'm not immediately sure to do this though since at this time we only have immutable access.

complexspaces (changes requested):
I like how this is looking so far, thank you for picking this up.

Cargo.toml
complexspaces: GitHub put my comment elsewhere by mistake, but I'm also in favor of this change. It also lines up with the `rustls` API for adding arbitrary roots to a config not being gated.

src/verification/others.rs
```diff
@@ -20,16 +20,35 @@ pub struct Verifier {
...
+            #[cfg(feature = "extra-roots")]
+            extra_roots: Vec::new(),
+        }
+    }
+
+    /// Creates a new verifier whose certificate validation is provided by
+    /// WebPKI, using root certificates provided by the platform and augmented by
+    /// the provided extra root certificates.
+    #[cfg(feature = "extra-roots")]
+    pub fn new_with_extra_roots(roots: impl Iterator<Item = rustls::OwnedTrustAnchor>) -> Self {
```
Ralith, line 45: Would it be better to use `impl IntoIterator` here?

src/verification/apple.rs
cpu: Copy-pasta err :spaghetti: :x: Thanks for catching.

src/verification/others.rs
cpu: I'll see if I can figure something out.

Cargo.toml
cpu: Removed the feature flag. Thanks for the input all :+1:

src/verification/android.rs
cpu: Tacked on an additional commit to implement this change.

src/verification/apple.rs
cpu: Fixed up.

src/lib.rs
cpu: Added some text to emphasize that `tls_config` uses safe defaults.

src/verification/others.rs
cpu: Switched, thanks!

cpu:
> I was thinking in terms of a specific API call that makes it trivial to add the webpki-roots specifically, but maybe the we can make that easy enough anyway by way of the pki-types project?

That seems like a good direction. It felt like for the purposes of platform-verifier it was better to be agnostic of the `webpki-roots` crate and just provide an option for an iterator of `OwnedTrustAnchors` of any provenance.

src/verification/others.rs
cpu: I'm not sure I see a path forward here :-/ Perhaps with some adjustments on the rustls side to not require owned representations of trust anchors?

@complexspaces Do you think this is blocking for the rest of the work?

complexspaces: I was thinking about this a little more, but couldn't we just wrap `extra_roots` in a `Mutex` and then use `Vec.drain` here? This shouldn't be in any kind of hot path since `init_verifier` is only called once per `Verifier` construction.

cpu: > couldn't we just wrap extra_roots in a Mutex and then use Vec.drain here?

That worked great, thank you. I had tried using a `Mutex` previously but I think I confused myself trying to use `get_mut` instead of `try_lock` and bumping into the same requirement of `mut self` (silly in retrospect).

I've used an `if let Ok(mut extra_roots) = self.extra_roots.try_lock() { ... }` construction in my last push. Do you think that's reasonable or should we handle the `Err` case differently? Maybe failing initialization?

complexspaces: Personally I would have just used `.unwrap()` because there's no way for the Mutex to get locked multiple times, so it could never be poisoned (which is the only route to a panic) but I'm fine with `if let Ok()` if others prefer that as long as we have a comment saying it can never fail.

cpu: > Personally I would have just used .unwrap() because there's no way for the Mutex to get locked multiple times, so it could never be poisoned (which is the only route to a panic)

That makes sense to me. I'll switch to `unwrap` and add a comment.

complexspaces (approved):

cpu:
@djc Would you like me to hold this for you to do another pass, or can I merge?

complexspaces:
FWIW the only other thing that came to mind is maybe allowing arbitrary roots to be added to any verifier for API consistency, but this is significantly complicated then adding them to the `webpki`-based verifier. Given that its probably best to wait for someone to present a use case  or PR for it.

djc:
I don't need to do another pass.

PR #28, opened 2023-08-30
verification: enable ip subj. validation on android.

### LINKED ISSUE #15: Support IP Address verification on Android
After the `webpki` upgrade inside of https://github.com/rustls/rustls-platform-verifier/pull/14, we now have the ability to verify IP address subject names on Android (Linux and WASM got this for free), once we upgraded.

The relevant ares are:
https://github.com/rustls/rustls-platform-verifier/blob/c18e4628dea5e48d11f9c06047d15d4862c38fc1/src/verification/android.rs#L283-L286

https://github.com/rustls/rustls-platform-verifier/blob/c18e4628dea5e48d11f9c06047d15d4862c38fc1/src/verification/android.rs#L191-L197

We should also make sure to enable the IP address portion of the mock test suite for Android at the same time to confirm the system supports the behavior.

cpu:
While the Android verifier defers to the platform verifier for most certificate validation options, it relies on `webpki` for ensuring that a certificate is valid for a given subject name. Since Webpki v0.100.x it's been possible to use `webpki` to validate IP address subject names, and so we want this capability to be enabled for the Android verifier.

This commit updates the Android `verifier` to accomplish this. Additionally it enables the mock IP address subject test cases in the mock test suite to ensure things work as expected.

In order to support this change the function signature of the inner `Verifier.verify_certificate` fn has to change from accepting a `&str` server name, to also accepting the `&rustls::ServerName` that the trait based `ServerCertVerifier.verify_server_cert` was already accepting. There are two main reasons for this:

1. If we try to pull out a `String` to pass forward from the `rustls::ServerName::IpAddress(&IpAddr)` using `IpAddr.to_string()` we'll get back a "compressed" address for IPv6 addresses. This is problematic when later trying to convert to a `webpki::IpAddrRef` for the validation call using `IpAddrRef::try_from_ascii_str`, because it doesn't support the compressed form.
2. By passing through the `rustls::ServerName` directly we can defer the actual process of interacting with `webpki` to the newly exposed [`rustls::client::verify_server_name` fn](https://docs.rs/rustls/latest/rustls/client/fn.verify_server_name.html) offered with the `dangerous_configuration` feature. This will ensure the logic for name validation is applied consistently between Rustls and the platform verifier. It also allows removing the Android specific `pki_name_error` helper, and the Android target's usage of the general `crate::verification::invalid_certificate` helper.

Resolves #15

### DIFF
--- a/src/tests/verification_mock/mod.rs
+++ b/src/tests/verification_mock/mod.rs
@@ -72,15 +72,11 @@ macro_rules! no_error {
 const ROOT1: &[u8] = include_bytes!("root1.crt");
 const ROOT1_INT1: &[u8] = include_bytes!("root1-int1.crt");
 const ROOT1_INT1_EXAMPLE_COM_GOOD: &[u8] = include_bytes!("root1-int1-ee_example.com-good.crt");
-#[cfg(any(windows, target_os = "macos", target_os = "linux"))]
 const ROOT1_INT1_LOCALHOST_IPV4_GOOD: &[u8] = include_bytes!("root1-int1-ee_127.0.0.1-good.crt");
-#[cfg(any(windows, target_os = "macos", target_os = "linux"))]
 const ROOT1_INT1_LOCALHOST_IPV6_GOOD: &[u8] = include_bytes!("root1-int1-ee_1-good.crt");
 
 const EXAMPLE_COM: &str = "example.com";
-#[cfg(any(windows, target_os = "macos", target_os = "linux"))]
 const LOCALHOST_IPV4: &str = "127.0.0.1";
-#[cfg(any(windows, target_os = "macos", target_os = "linux"))]
 const LOCALHOST_IPV6: &str = "::1";
 
 #[cfg(any(test, feature = "ffi-testing"))]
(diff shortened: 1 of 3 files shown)


src/verification/android.rs
```diff
@@ -73,7 +72,7 @@ impl Verifier {
         &self,
         end_entity: &rustls::Certificate,
         intermediates: &[rustls::Certificate],
-        server_name: &str,
+        server_name: &rustls::ServerName,
```
djc, line 75: If we could make this method private we could just take a `&str` here and do the conversion to string in the `ServerCertVerifier::verify_server_cert()` method, instead? That would avoid the `unreachable!()` shenanigans, at least...

cpu: I already made this `fn` private in https://github.com/rustls/rustls-platform-verifier/pull/24 so I think its fine to bring that change in here.

We'd need to pass the `rustls::ServerName` in addition to the `&str`, but otherwise seems OK.

cpu: > If we could make this method private we could just take a &str here

Also, I think it needs to be `String`. You can get a `&str` from `DnsName` but not from `IpAddr` AFAICT.

cpu: Pushed an update to accomplish the above. PTAL

djc (approved):

src/tests/verification_mock/mod.rs
```diff
@@ -72,15 +72,35 @@ macro_rules! no_error {
 const ROOT1: &[u8] = include_bytes!("root1.crt");
 const ROOT1_INT1: &[u8] = include_bytes!("root1-int1.crt");
 const ROOT1_INT1_EXAMPLE_COM_GOOD: &[u8] = include_bytes!("root1-int1-ee_example.com-good.crt");
-#[cfg(any(windows, target_os = "macos", target_os = "linux"))]
+#[cfg(any(
+    windows,
+    target_os = "android",
+    target_os = "macos",
+    target_os = "linux"
+))]
```
complexspaces, line 80: At this point, I believe we're now exactly matching the `cfg` guard for the entire file itself:
```

#![cfg(any(
    windows,
    target_os = "android",
    target_os = "macos",
    target_os = "linux"
))]
```

Given that future iOS and WASM testing would use an existing, fully-supported, verifier, I don't anticipate needing to re-guard these. Could we try removing all of these 4-platform `#[cfg]` guards and the ones inside the test generation macro for simplicity?

src/verification/android.rs
```diff
@@ -189,17 +189,10 @@ impl Verifier {
...
-                        );
-                        if cert.verify_is_valid_for_subject_name(name).is_ok() {
-                            Ok(())
-                        } else {
-                            Err(InvalidCertificate(CertificateError::NotValidForName))
-                        }
+                        rustls::client::verify_server_name(
+                            &rustls::server::ParsedCertificate::try_from(end_entity)?,
+                            server_name,
+                        )
```
complexspaces, line 195: This is much nicer, thanks for calling this method out!

src/verification/android.rs
```diff
@@ -280,8 +265,11 @@ impl ServerCertVerifier for Verifier {
...
 
-        let server = match server_name {
-            rustls::ServerName::DnsName(name) => name.as_ref(),
+        // Verify the server name is one that we support and extract a string to use
+        // for the platform verifier call.
+        let server_name_str = match server_name {
+            ServerName::DnsName(dns_name) => dns_name.as_ref().to_string(),
+            ServerName::IpAddress(ip_addr) => ip_addr.to_string(),
             _ => return Err(unsupported_server_name()),
         };
```
complexspaces, line 274: Suggestion: We could use a late binding here to avoid the extra hot-path allocation. We did the same thing in the the Apple and Windows verifier as well: https://github.com/rustls/rustls-platform-verifier/blob/1a5250c61d30fbac4af26c4de07561755359b4f6/src/verification/apple.rs#L194-L202

https://github.com/rustls/rustls-platform-verifier/blob/1a5250c61d30fbac4af26c4de07561755359b4f6/src/verification/windows.rs#L608-L614

complexspaces (changes requested):
This ended up much easier then I expected based on when I last tried this, so this is great to see. I wonder if your first point about compressed IPv6 addresses was what was messing this up in the last attempt, since I think that I did try roundtripping it through string parsing.

src/tests/verification_mock/mod.rs
cpu: > Could we try removing all of these 4-platform #[cfg] guards 

Good call, thanks! I was worried about the future iOS/WASM testing but as you mention that probably isn't a good reason to keep these.

> and the ones inside the test generation macro for simplicity?

I left these because the "stapled_revoked" test cases still need to be guarded to not run on Linux.

src/verification/android.rs
cpu: >  We could use a late binding here

Ohhhh! So clever. I'm not sure I realized this was possible :sweat_smile: It neatly resolves the issue I had where I couldn't return a ref to an intermediate var scoped within the match.

Thanks for the suggestion :+1:

src/tests/verification_mock/mod.rs
complexspaces: Ah ok, that makes sense 👍

complexspaces (approved):
This looks good to go now, LGTM.

cpu:
Thanks for all the help :-)

Issue #26, opened 2023-08-29
GitHub CI using deprecated actions


cpu:
The CI workflows are currently using `actions-rs` tasks that are [unmaintained](https://github.com/actions-rs/toolchain/issues/216) and affected by GitHub's deprecation of Node 12.

> The following actions uses node12 which is deprecated and will be forced to run on node16: actions-rs/toolchain@v1, actions-rs/cargo@v1. For more info: https://github.blog/changelog/2023-06-13-github-actions-all-actions-will-run-on-node16-instead-of-node12-by-default/

The following workflow task need to be replaced:

* `actions-rs/toolchain@v1` used by `ci.yaml`
* `actions-rs/cargo@v1` used by `ci.yaml`
* `actions-rs/audit-check@v1` used by `audit.yml`

In the other Rustls projects we've switched to https://github.com/dtolnay/rust-toolchain, but it doesn't offer an equivalent to `audit-check`. That one might require a separate 3rd party action.

complexspaces:
`audit-check` is small enough we could probably utilize GitHub Actions cache and a manual `cargo-install` to accomplish the same thing. I'm in favor of migrating the other ones to Dtolnay's.

djc:
What does audit-check do? Can we just pick up cargo-deny? It consumes the advisory database and helps with a bunch of other stuff, too.

complexspaces:
I see `cargo-deny` as a superset of `cargo-audit`, so I don't see an issue with that if we have an easy way to utilize it in CI.

djc:
#29 replaces audit-check with cargo-deny.

PR #29, opened 2023-09-01
Use cargo-deny for auditing

djc:
Also update the lockfile, since there is a vulnerability in the version of h2 that was being used.

(Not sure why audit-check hadn't caught that?)

See #26 for discussion on why actions-rs/audit-check needs to be replaced.

### DIFF
--- a/deny.toml
+++ b/deny.toml
@@ -1,0 +1,15 @@
+[advisories]
+unsound = "deny"
+yanked = "deny"
+unmaintained = "deny"
+
+[licenses]
+allow-osi-fsf-free = "either"
+copyleft = "warn"
+allow = ["MPL-2.0"]
+exceptions = [{ allow = ["ISC", "MIT", "OpenSSL"], name = "ring" }]
+
+[[licenses.clarify]]
+name = "ring"
+expression = "ISC AND MIT AND OpenSSL"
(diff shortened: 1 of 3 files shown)


djc:
Ah, it looks like the audit workflow was disabled because of inactivity (which GitHub does automatically for scheduled workflows).

deny.toml
```diff
@@ -0,0 +1,15 @@
+[advisories]
+unsound = "deny"
+yanked = "deny"
+unmaintained = "deny"
+
+[licenses]
+allow-osi-fsf-free = "either"
+copyleft = "warn"
+allow = ["MPL-2.0"]
```
djc, line 9: `MPL-2.0` is considered `copyleft` but as far as I can tell it avoids the issues associated with viral licenses like the GPL. @complexspaces do you have some kind of policy on what licenses are allowed in 1P dependencies?

cpu (approved):
Thanks for picking this up. Looks good to me. Will hold merge for ComplexSpaces to +1.

deny.toml
complexspaces: We do, but we made the same observation about the MPL, so it is on our internal allow-list. I don't see any issues with allowing it here either 👍.

complexspaces (approved):
This looks good to me as well, thanks for jumping on this.

PR #31, opened 2023-09-02
Add Android release artifact stripping and verification

### LINKED ISSUE #5: Explicitly remove and/or verify Android mock test code in release builds
In the Android verifier implementation, [we have several](https://github.com/rustls/rustls-platform-verifier/blob/b25faa4a2b9afb1e2c8f5d883fd9d8d91972519f/android/rustls-platform-verifier/src/main/java/org/rustls/platformverifier/CertificateVerifier.kt#L95-L159) class fields and conditionals that are solely in place to facilitate testing with mocked certificates and roots. While the use of said store is [guarded](https://github.com/rustls/rustls-platform-verifier/blob/b25faa4a2b9afb1e2c8f5d883fd9d8d91972519f/android/rustls-platform-verifier/src/main/java/org/rustls/platformverifier/CertificateVerifier.kt#L220-L241), it would be better if it was provable these couldn't affect a production app, similar to `#[cfg(debug_assertions)]` in Rust.

I can think of two initial approaches for this:
- Create custom Proguard rules that remove this code from anything but an explicit integration test build
- Automatically verify that the code doesn't end up being loaded at runtime

complexspaces:
This PR's goal is to make sure we are not shipping any test-only code to end users, both for security and code size reasons. Unlike Rust where compile-time code elimination is a first-class feature, more work is required to obtain the same result in Java/Kotlin. In total, there are two parts of making this work:
- Ensuring the Java compiler can remove branches which access test-only methods and fields on the verifier class.
    - We accomplish this by introducing, and switching to, a proper constant into the Android library's `BuildConfig` which can be evaluated at bytecode compile time.
    - With this new true constant, the Java compiler is able to delete our test branches reliably.
- Configuring Proguard/R8 to run on the resulting release artifact to strip out all of the now-unused test-related code.

Additionally, a script was added that can verify the contents of the current artifact configuration. It's also been added to CI here for automatic verification.

To visually confirm this, you can either open the decompiled files generated by the new script in your editor of choice, or you can use a visual Java decompiler. For this PR I used both `jadx-gui` and Android Studio's built-in decompiler to confirm what class parts were present. For example, you can see that every `*mock*` class part is not present anymore:
image

Closes #5

### DIFF
--- a/ci/verify_android_release.sh
+++ b/ci/verify_android_release.sh
@@ -1,0 +1,41 @@
+#!/usr/bin/env bash
+
+# This script's purpose is to verify that no test-only code is present inside of the release-mode Android artifact.
+# It is validating that `javac` is performing the dead-code elimiation we expect and that `proguard` is deleting the
+# unreferenced test code. This can be ran both locally and in CI.
+#
+# It accomplishes this goal by building the artifact and then running a decompiler on it to look for names we expect or do not.
+
+set -euo pipefail
+
+mkdir -p ./android/verification
+
+pushd ./android/
+
+./gradlew clean
+./gradlew assembleRelease
+
+pushd ./verification
+
+if [ ! -f "./bin/jadx" ]; then
+    echo "Decompiler not yet installed, downloading jadx"
+    curl -L https://github.com/skylot/jadx/releases/download/v1.4.7/jadx-1.4.7.zip --output jadx.zip
+    unzip jadx.zip
+    echo "jadx downloaded"
+fi
+
+./bin/jadx -d decompiled ../rustls-platform-verifier/build/outputs/aar/rustls-platform-verifier-release.aar
+
+if grep -r -q "mock" ./decompiled; then
+    echo "❌ Test-only code exists in release artifact! Please review changes made to locate the cause".
+    exit 1
+else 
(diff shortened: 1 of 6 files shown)


ci/verify_android_release.sh
```diff
@@ -0,0 +1,41 @@
+#!/usr/bin/env bash
+
+# This script's purpose is to verify that no test-only code is present inside of the release-mode Android artifact.
+# It is validating that `javac` is performing the dead-code elimiation we expect and that `proguard` is deleting the
+# unreferenced test code. This can be ran both locally and in CI.
+#
+# It accomplishes this goal by building the artifact nd then running a decompiler on it to look for names we expect or do not.
```
cpu, line 7: tiny typo:
```suggestion
# It accomplishes this goal by building the artifact and then running a decompiler on it to look for names we expect or do not.
```

cpu (approved):
Nice work! I was able to use the verification script locally with no trouble and could reproduce your manual testing results.

djc (approved):

ci/verify_android_release.sh
complexspaces: Thanks!

PR #32, opened 2023-09-08
Introduce new Android artifact distribution strategy 

complexspaces:
The goal of this pull request is to resolve the outstanding issues with the current Android component distribution system and provide a new one that checks as many boxes as possible without causing issues elsewhere.

To accomplish this, a new method has been built to completely replace the old way. For the exact details, see the documentation included in the changes. At a high level, we've switched to distributing pre-compiled versions of the Android component via a Maven local repository. A new crate, `rustls-platform-verifier-android`, is responsible for containing these artifacts. Upon downloading this new crate via `cargo`, Gradle is able to seamlessly locate the files inside Cargo's dependency cache and compile the packaged AAR into the main application's build.

This approach eliminates the toolchain synchronization problems and poor Gradle workarounds that affected the last approach. Additionally, it has minimal-to-no affect on Gradle build or configuration performance and fits in much better with its compilation model.

### Further Improvements

See https://github.com/rustls/rustls-platform-verifier/pull/33, which builds upon this packaging logic to provide always-available Android artifacts based off the current `main`. The idea behind this is to attempt to preserve the convenience of the Git dependencies that `cargo` supports. Depending on the release cadence here in the future, people may need to depend on hot-fixes for short periods. By hosting these pre-compiled artifacts in a dedicated, opt-in branch, those with Android apps can add a Git dependency on `main-with-maven` instead to avoid the need to temporarily vendor AAR artifacts.


rustls-platform-verifier/Cargo.toml
```diff
@@ -0,0 +1,76 @@
+[package]
```
djc, line 1: Could you get git to recognize this as a move/rename?

README.md
```diff
@@ -93,6 +93,13 @@ new releases of `rustls-platform-verifier` are published. If you only use publis
 These script snippets can be tweaked as best suits your project, but the `cargo metadata` invocation must be included so that the Android
 implementation part can be located on disk.
 
+If your Android application makes use of Proguard for optimizations, its important to make sure that the Android verifier component isn't optimized 
+out because it looks like dead code". Proguard is unable to see any JNI usage, so your rules must manually opt into keeping it. THe following rule
```
djc, line 97: Stray `"` near "dead code".

djc (approved):
LGTM, nice work!

admin/RELEASING.md
```diff
@@ -0,0 +1,18 @@
...
+
+1. Update main crate'a version in `rustls-platform-verifier/Cargo.toml`, and in any additional places.
+2. If any non-test changes have been made to the `android` directory since the last release:
+    1. Update Android artifact version in `android-release-support/Cargo.toml`
+    2. Bump dependency version of the Android support crate in `rustls-platform-verifier/Cargo.toml` to match the new one
+    3. Commit version increase changes on the release branch
+    4. Run `ci/package_android_release.sh` in a UNIX compatible shell
+    5. (Optional) `cargo publish -p rustls-platform-verifier-android --dry-run`
+    6. (Optional) Inspect extracted archive to ensure the local Maven repository artifacts are present
+    7. Publish the Android artifacts' new version: `cargo publish -p rustls-platform-verifier-android`
```
complexspaces, line 16: @djc @cpu One open question I have here is about the Git tagging/release process: Should we bother attaching the precompiled artifacts to any tag or GitHub release? This would let someone easily download/verify it if they desired. Alternatively, they can just download it from crates.io and do the same thing.

If we do want to include it here, I haven't really come up with a good order to do so.

rustls-platform-verifier/Cargo.toml
complexspaces: I can give it a try, I'm not really sure why it didn't already. Maybe because the original filename is still there?

admin/RELEASING.md
djc: Letting them download from crates.io seems fine for now, can always iterate in response to feedback.

complexspaces: That seems reasonable to me. Resolving.

rustls-platform-verifier/Cargo.toml
complexspaces: The only way I was able to accomplish this was by entirely removing the original `Cargo.toml` in the first commit and then adding the workspace contents in a second. It does make this set of commits less atomic, but in exchange the history should be preserved. Thoughts?

admin/RELEASING.md
```diff
@@ -0,0 +1,18 @@
...
+This document records the steps to publish new versions of the crate since it requires non-trivial preperation and ordering
+that needs to be remembered due to the Android component's distribution.
+
+## Steps
+
+1. Update main crate'a version in `rustls-platform-verifier/Cargo.toml`, and in any additional places.
+2. If any non-test changes have been made to the `android` directory since the last release:
+    1. Update Android artifact version in `android-release-support/Cargo.toml`
+    2. Bump dependency version of the Android support crate in `rustls-platform-verifier/Cargo.toml` to match the new one
+    3. Commit version increase changes on the release branch
```
cpu, line 12: Do you have any preferences about how to create/name the release branch? We use [rel-xxx](https://github.com/rustls/rustls/blob/main/RELEASING.md#maintenance-point-releases) in rustls/rustls and rustls/webpki fwiw

admin/RELEASING.md
```diff
@@ -0,0 +1,18 @@
...
+    4. Run `ci/package_android_release.sh` in a UNIX compatible shell
+    5. (Optional) `cargo publish -p rustls-platform-verifier-android --dry-run`
+    6. (Optional) Inspect extracted archive to ensure the local Maven repository artifacts are present
```
cpu, line 15: I think it would be helpful to provide commands to run for this too.

admin/RELEASING.md
```diff
@@ -0,0 +1,18 @@
...
+
+## Steps
+
+1. Update main crate'a version in `rustls-platform-verifier/Cargo.toml`, and in any additional places.
+2. If any non-test changes have been made to the `android` directory since the last release:
+    1. Update Android artifact version in `android-release-support/Cargo.toml`
+    2. Bump dependency version of the Android support crate in `rustls-platform-verifier/Cargo.toml` to match the new one
+    3. Commit version increase changes on the release branch
+    4. Run `ci/package_android_release.sh` in a UNIX compatible shell
+    5. (Optional) `cargo publish -p rustls-platform-verifier-android --dry-run`
```
cpu, line 14: When I run this immediately after `ci/package_android_releas.sh` it bails from a dirty worktree:
```
❯ cargo publish -p rustls-platform-verifier-android --dry-run
    Updating crates.io index
error: 4 files in the working directory contain changes that were not yet committed into git:

maven/rustls/rustls-platform-verifier/0.1.0/rustls-platform-verifier-0.1.0.pom
maven/rustls/rustls-platform-verifier/0.1.0/rustls-platform-verifier-0.1.0.aar
maven/rustls/rustls-platform-verifier/maven-metadata-local.xml
maven/pom.xml

to proceed despite this and include the uncommitted changes, pass the `--allow-dirty` flag
```

Is the releaser supposed to use `--alow-dirty` since the include directive packages these files for a real release?

android-release-support/src/lib.rs
```diff
@@ -0,0 +1,71 @@
+//! # rustls-platform-verifier-android
+//!
+//! This crate is an implementation detail of the actual [rustls-platform-verifier](https://github.com/rustls/rustls-platform-verifier) crate.
+//!
+//! It contains no Rust code and is solely intended as a convenient delivery mechanism for the supporting Kotlin code that the main crate
+//! requires to perform TLS certificate validation using Android's APIs.
+//!
+//! Other crates should not directly depend on this crate in any way, as nothing about it is considered stable and probably useless elsewhere.
```
cpu, line 8: small typo:

> as nothing about it is considered stable and probably useless elsewhere.

:arrow_down: 

> as nothing about it is considered stable and it is probably useless elsewhere.

android-release-support/src/lib.rs
```diff
@@ -0,0 +1,71 @@
...
+//!
+//! ## Details
+//!
+//! Note: Everything in this section is subject to change at any time. Semver may not be followed either.
```
cpu, line 12: editorial nit:
> Semver may not be followed either.

:arrow_down: 

> Semver may not be followed.

android-release-support/src/lib.rs
```diff
@@ -0,0 +1,71 @@
...
+//!
+//! ### Why?
+//!
+//! It was the best middle ground between several tradeoffs. The important ones, in priority order, are:
+//! - Automatically keeping component versions in sync
+//! - Allowing well-tested and well-known `cargo` dependency management patterns to apply everywhere
+//! - Providing a smooth developer experience as an Android consumer of `rustls-platform-verifier`
+//!
+//! Firstly, what alternatives are available for distributing the component? The other two known are source distribution n some form (here, it will be through crates.io)
```
cpu, line 21: small typo:

> are source distribution n some form

:arrow_down: 

> are source distribution in some form

android-release-support/src/lib.rs
```diff
@@ -0,0 +1,71 @@
...
+//! and Maven Central. Starting with the first, its become infeasible due to toolchain syncing requirements. If the Android component is
+//! built as part of the host app's Gradle build, then it becomes subject to any Gradle or AGP incompatibilites/requirements. In practice this means
```
cpu, line 23: WDYT about expanding AGP for the first usage to Android Gradle Plugin (?)

android-release-support/src/lib.rs
```diff
@@ -0,0 +1,71 @@
...
+//! could be done with raw JNI calls, but this would easily be 3x the size of the existing implementation and require huge amounts of `unsafe`
+//! to review then audit.
+//!
+//! ### The solution
+//!
+//! The final design was built to avoid the pitfalls the previous two options mentioned. To build it, we rely on CI and packaging scripts to build
+//! the Android component into a prebuilt AAR file before creating a release. Next, a [on-disk Maven repository](https://maven.apache.org/repositories/local.html)
+//! is hosted inside of this repository. Only the unchanging file structure of it is kept checked-in, to avoid churn. The remaining parts are filled in
+//! during the packaging/release process, before being included in `cargo package` via an `include` Cargo.toml directive. Finally, once the repository has had
+//! its artifacts added, this crate is published to crates.io containing it. Then, the main crate ensures its downloaded when an Android target is compiled for via
```
cpu, line 45: editorial:

> Finally, once the repository has had its artifacts added, this crate is published to crates.io containing it.

:arrow_down: 

> Finally, once the repository has had its artifacts added the crate containing the Maven repository is published to crates.io.

cpu, line 45: small typo:
> Then, the main crate ensures its downloaded when an Android target is compiled for via a platform-specific dependency.

:arrow_down: 

>  Then, the main crate ensures it's downloaded when an Android target is compiled via a platform-specific dependency.

android-release-support/src/lib.rs
```diff
@@ -0,0 +1,71 @@
...
+//! a platform-specific dependency.
+//!
+//! On the Gradle side, we include a very small snippet of code for users to include in their `settings.gradle` file to dyanmically locate the local maven repository
```
cpu, line 48: WDYT about linking directly to that Gradle snippet from here.

admin/RELEASING.md
```diff
@@ -0,0 +1,18 @@
+# How-to release `rustls-platform-verifier`
```
cpu, line 1: WDYT about adding some of the text from the Rustls [RELEASING.md](https://github.com/rustls/rustls/blob/main/RELEASING.md#maintenance-point-releases) that talks about proposing release notes?

Do you think using GitHub releases would also be helpful? I think it's a good way to help the many Dependabot users.

ci/package_android_release.sh
```diff
@@ -0,0 +1,38 @@
...
+
+set -euo pipefail
+
+if ! type mvn > /dev/null; then
+  echo "The maven CLI, mvn, is required to run this script."
+  echo "Download it from: https://maven.apache.org/download.cgi"
+  exit 1
+fi
+
+version=$(cat android-release-support/Cargo.toml | grep -m 1 "version = " | tr -d "version= " | tr -d '"')
```
cpu, line 16: shellcheck nit: we can drop the `cat` here and use `grep` directly to save a process:

```bash
version=$(grep -m 1 "version = " android-release-support/Cargo.toml | tr -d "version= " | tr -d '"')
```

ci/package_android_release.sh
```diff
@@ -0,0 +1,38 @@
...
+
+pushd ./android-release-support
+
+artifact_path="../android/rustls-platform-verifier/build/outputs/aar/$artifact_name"
+
+# Ensure no prior artifacts are present
+git clean -dfX "./maven/"
+
+cp ./pom-template.xml ./maven/pom.xml
+sed -i "" "s/\$VERSION/$version/" ./maven/pom.xml
```
cpu, line 36: fwiw on my system w/ GNU sed 4.9 this bailed with the error:
```
sed: can't read s/$VERSION/0.1.0/: No such file or directory
```

It needed to be changed to remove the space between `-i` and `""`:

```bash
sed -i"" "s/\$VERSION/$version/" ./maven/pom.xml
```

cpu (approved):
very nice work! Thank you for the detailed descriptions of why this option makes the most sense. Especially the parts about the discarded alternatives. It's very helpful background.

I flagged a handful of typos and some minor scripting things but none feel blocking.

ci/package_android_release.sh
complexspaces: I noticed this as well when working on the automation PR. I'll cherry-pick the fix I made over there to this branch. There's an annoying incompatibility between macOS's `sed` and GNU's `sed`.

admin/RELEASING.md
```diff
@@ -0,0 +1,18 @@
...
+
+This document records the steps to publish new versions of the crate since it requires non-trivial preperation and ordering
+that needs to be remembered due to the Android component's distribution.
+
+## Steps
+
+1. Update main crate'a version in `rustls-platform-verifier/Cargo.toml`, and in any additional places.
```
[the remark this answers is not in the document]
complexspaces: I don't think there are any today, I only added this for future-proofness. I can remove it for now if you think that would be clearer.

admin/RELEASING.md
complexspaces: I don't have any preference, no. In other open source repositories for 1P I use `release-xxx` and then delete it once the tag has been created.

admin/RELEASING.md
complexspaces: They shouldn't need to, no. I'm not sure what the difference is yet but the new gitignore rules _should_ be preventing these from making the worktree dirty. Instead, the `include` directive in the Android support crate will make sure to grab them despite not being part of Git and remove the need for `--allow-dirty`.

admin/RELEASING.md
complexspaces: I'm in favor of using GitHub releases. I omitted a bit from these exact steps but in other repositories I maintain I use GitHub releases to create both a release and tag at the same time. After creating the tag via release, I check it out and run `cargo publish`. I'll try adding more specifics here.

rustls-platform-verifier/Cargo.toml
djc: Probably not worth it.

admin/RELEASING.md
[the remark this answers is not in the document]
cpu: > I can remove it for now if you think that would be clearer.

yeah, let's remove it for now. I'd rather try to update this when there are concrete cases to discuss if it's all the same to you.

admin/RELEASING.md
cpu: > I don't have any preference, no. 

Let's standardize on the Rustls approach then to have one less thing to map between repos.

> and then delete it once the tag has been created.

We've been leaving our release branches around for subsequent point releases to use, but I don't feel strongly.

admin/RELEASING.md
cpu: >  Running the script doesn't change git status for me locally.

It's the same for me FWIW. 

> They shouldn't need to, no. I'm not sure what the difference is yet but the new gitignore rules should be preventing these from making the worktree dirty.

It sounds like this might be intentional behaviour on the part of `cargo`: https://github.com/rust-lang/cargo/issues/12294

ci/package_android_release.sh
cpu: > There's an annoying incompatibility between macOS's sed and GNU's sed.

argh, of course there is :dizzy_face:

complexspaces: This should be fixed now.

admin/RELEASING.md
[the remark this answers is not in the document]
complexspaces: Done!

android-release-support/src/lib.rs
```diff
@@ -0,0 +1,71 @@
...
+//! It was the best middle ground between several tradeoffs. The important ones, in priority order, are:
+//! - Automatically keeping component versions in sync
+//! - Allowing well-tested and well-known `cargo` dependency management patterns to apply everywhere
+//! - Providing a smooth developer experience as an Android consumer of `rustls-platform-verifier`
+//!
+//! Firstly, what alternatives are available for distributing the component? The other two known are source distribution in some form (here, it will be through crates.io)
+//! and Maven Central. Starting with the first, its become infeasible due to toolchain syncing requirements. If the Android component is
+//! built as part of the host app's Gradle build, then it becomes subject to any Gradle or Android Gradle Plugin incompatibilites/requirements. In practice this means
+//! the AGP version between this project and the main application have to match all the time. Sometimes this works, but it becomes challenging/unfeasible
+//! during yearly toolchain/SDK upgrades and is not maintainable long term. Note that this is the _only_ option in this section which retains compatible
```
tshepang, line 25:
```suggestion
//! during yearly toolchain/SDK upgrades and is not maintainable long term. Note that this is the _only_ option in this section which retains compatibility
```

android-release-support/src/lib.rs
```diff
@@ -0,0 +1,71 @@
...
+//! For some, the notion of shipping something pre-compiled with an existing source distribution might seem incorrect, or insecure. However in this specific case,
+//! putting aside the fact shipping Kotlin code doesn't work (see above), there are many reasons this isn't the case:
+//! - Shipping pre-compiled artifacts is normal in the Java ecosystem. Maven Central and other package repositories do the same thing and serve `.jar` downloads.
+//! - Those not using Android will never download the pre-compiled AAR file.
+//! - The artifacts are incredibly easy to reproduce given an identicial compilation toolchain.
+//! - The artifacts are not native executables, or raw `.jar` files, so they can't be accidentally executed on a host system.
+//!
+//! ## Summary
+//!
+//! In summary, the selected distribution method avoids most of the previous pitfalls while still balancing a good experience for `cargo` annd Gradle users. Some of its
```
tshepang, line 64:
```suggestion
//! In summary, the selected distribution method avoids most of the previous pitfalls while still balancing a good experience for `cargo` and Gradle users. Some of its
```

rustls-platform-verifier/Cargo.toml
```diff
@@ -0,0 +1,69 @@
...
+name = "rustls-platform-verifier"
+version = "0.1.0"
+authors = ["ComplexSpaces <complexspacescode@gmail.com>", "1Password"]
+description = "rustls-platform-verifier supports verifying TLS certificates in rustls with the operating system verifier"
+keywords = ["tls", "certificate", "verification", "os", "native"]
+repository = "https://github.com/1Password/rustls-platform-verifier"
```
tshepang, line 7:
```suggestion
repository = "https://github.com/rustls/rustls-platform-verifier"
```

cpu:
@complexspaces WDYT about rebasing this and merging it as-is (maybe modulo the small fixes called out by tshepang)? I think my feedback about more documentation for the release process doesn't need to be blocking. We could iterate on those parts based on experience.

cpu:
@complexspaces small bump. Would it be helpful if I rebased this branch and addressed tshepang's comments?

complexspaces:
Hi, sorry for the delay @cpu. I've once again gotten behind on things. I will address tshepang's comments this week and try to clean up some of the other smaller comments as well before Monday next week. At that point, we can evaluate if this looks good enough for an initial publishing attempt?

admin/RELEASING.md
complexspaces: I will try to investigate this more, but do you think this is a blocker? The Android-related artifacts are easily reproducible so my hope is that anyone who is less comfortable with "binary" artifacts being included in a crate's contents would be able to checkout the repository from the Git tag, regenerate these artifacts and assert their checksums match.

cpu:
> I will address tshepang's comments this week and try to clean up some of the other smaller comments as well before Monday next week.

Great, thank you! I know it's a busy time of year :-)

>  At that point, we can evaluate if this looks good enough for an initial publishing attempt?

Sounds good to me. I'm eager to get something out into the world before the new year if we can convince ourselves that all of our ducks are in a row. It might be nice to land https://github.com/rustls/rustls-platform-verifier/pull/42 but it seems unlikely that Reqwest and the other blocking dependencies will be ready on a short timescale. The very recent Hyper 1.0 update layers a bit more complication into the mix.

admin/RELEASING.md
cpu: I don't think it needs to be considered a blocker :+1:

cpu:
Checking in again: anything I can do to help move closer to publishing a release?

cpu:
Here's [an updated branch](https://github.com/cpu/rustls-platform-verifier/tree/cpu-android-artifact-distribution_dev) that's been rebased on `main` and had the majority of the outstanding review feedback addressed.

complexspaces:
I believe that I've finished this branch and it's ready for another review. It took a bit longer then expected (1-2 days last week), and I apologize for the inconsistency. I ended up needing to port the `.gradle` script examples to Kotlin in order to end-to-end test the release/dependency process. 1Password internally moved to `.gradle.kts` files, and that was also my "test user" project :)

In the extra time I also discovered some Gradle performance/correctness improvements as well so IMO the initial release will be more polished as a result.

New changes include:
- Fixing up the typos that were pointed out, and more I noticed after temporarily enabling a spellchecker.
- Cherry-picking in and then squashing @cpu's release documentation suggestions.
- Resolving the last of the threads I believed were important (notably the context around `--allow-dirty`, and documenting a future TODO that may remove the need for it so no one forgets)
- Updating the Gradle integration example to remove the hacky `Memoize` usage and instead place the Maven repository definition in the right place.
    - This saves on configuration time and ensures the script is only run once the Gradle-intended way 😅.
- Documented intended Gradle sync performance/behavior so regressions have a baseline to report against.
- Added steps for verifying the `rustls-platform-verifier-android` dry-run package both on the filesystem and inside a Gradle project for testing.

Using all of the now-documented steps, I tested a few things to build confidence in the first release we make: 
- The Gradle integration is working as expected WRT caching and version updates (using a filesystem `rustls-platform-verifier` dependency patch)
- Filesystem crate patches are recognized for local development/overrides in both Cargo and Gradle.
- The extracted dry-run crate package is recognized by Gradle as well.

cpu (approved):
Thank you! The latest revisions look good to me :rocket:

PR #33, opened 2023-09-09
WIP: Automated android branch builds

complexspaces:
This PR is a followup to #32, in order to reduce friction around Cargo Git dependencies.

To create this, these changes introduce a new GitHub Action workflow that runs on every successful merge to main. The action follows these high-level steps:
1.  Reset the `main-with-maven` branch state to match `main`.
2. Run the packaging script and generate the local maven repository
3. Run `git add --force` to include the generated files and artifacts.
4. Commit the artifacts in a dedicated commit that is clear in purpose.
5. Force push and overwrite the `main-with-maven` branch with the new artifacts.

As the end result, there is a branch with identical contents to `crates.io` releases. As noted in the parent PR, this allows us to keep some of the nice properties of Cargo's Git dependency management and avoid friction between true releases.

One open question is if this should wipe out `main-with-maven`'s Git history each time or if it should perform a `git merge` instead. The upsides of the former are that its simple, clean over time, and easy to follow. The downside is that a merge to `main` may render that commit revision useless for those. It should be available for quite some time though unless [GitHub intervenes](https://github.com/orgs/community/discussions/23262#discussioncomment-3239760) and manually purges the repository cache. I lean towards wiping it every time given the above.


complexspaces:
This will need some cleanup to actually checkout the correct branch and push only the new commit, but that is blocked on the parent PR merging so `main` has the new packaging script and template available.

With that said, this confirms the general idea works. You can see a commit with the pre-built maven repository here: https://github.com/rustls/rustls-platform-verifier/commit/875f270a4d2e5d29f3988ddc425c126e0fd90595

.github/workflows/android_builds.yml
```diff
@@ -0,0 +1,52 @@
...
+    #   - closed
+
+name: Update Android Artifacts
+
+permissions:
+  contents: write
+
+jobs:
+  update-android-branch:
+    # if: github.event.pull_request.merged == true
```
cpu, line 15: Tag for fixup.

.github/workflows/android_builds.yml
```diff
@@ -0,0 +1,52 @@
...
+        git checkout -b main-with-maven
+        # git reset --hard origin/main
+        git reset --hard origin/automated-android-branch
+
+        echo "Building new Android release from main"
+        ./ci/package_android_release.sh
+
+        git add --force android-release-support/*
+        git commit -am "[Automated] Bundle Android component artifacts"
+        # git push -f origin main-with-maven
```
cpu, line 51: one more pre-merge fixup

.github/workflows/ci.yml
```diff
@@ -3,10 +3,10 @@ on:
...
       - main
       - "*_dev"
-  pull_request:
-  merge_group:
-  schedule:
-    - cron: '0 18 * * *'
+  #pull_request:
+  # merge_group:
+  #schedule:
+  #  - cron: '0 18 * * *'
```
cpu, line 9: Is there a reason to disable these? Leftover from testing?

.github/workflows/android_builds.yml
```diff
@@ -0,0 +1,52 @@
...
+  pull_request:
+    branches:
+      - main
+    # types:
+    #   - closed
+
+name: Update Android Artifacts
+
+permissions:
+  contents: write
```
cpu, line 11: Maybe prudent to leave a warning at the top of this YML that changes to this workflow should be carefully vetted since it can write to the repository.

I wonder if branch protection rules to require pull-requests into `main` would protect against this workflow pushing `main` or if that would be possible through malice or bugs in the script :-/

cpu (commented):
Cool stuff, thank you :-)

I'm a little bit nervous about GitHub workflows that use the `$GITHUB_TOKEN` to write back to the repository, but I think the risk/reward tradeoff is sensible here and we can approach changes with care.

If there's a way to use branch protection rules (or some other mechanism) to further limit the workflow to only being able to operate with the Android maven branches we intend that would be a nice extra addition :crossed_fingers:

.github/workflows/android_builds.yml
complexspaces: I believe branch protection rules do apply to Workflow Actions as well, although I'm unsure we can exactly test the existing `main` protection rule without risking a little bit.

Following this comment from you:
> If there's a way to use branch protection rules (or some other mechanism) to further limit the workflow to only being able to operate with the Android maven branches we intend that would be a nice extra addition 🤞

I tried adding a custom ruleset (see [here](https://github.com/rustls/rustls-platform-verifier/settings/rules/88258)) to the repository that applies to every branch except `main-with-maven`. I then tried to push from the CI job to a testing branch. [It failed to push](https://github.com/rustls/rustls-platform-verifier/actions/runs/6132043096/job/16642549893?pr=33) with a permissions error like we'd want.

The downside of it seems to be that it treats our pushes to working branches in the repository as rule bypasses as well, which causes extra noise locally 😞
```
remote: Bypassed rule violations for refs/heads/automated-android-branch:
remote:
remote: - Cannot update this protected ref.
remote:
To github.com:rustls/rustls-platform-verifier.git
   036ac7e..e147cb2  automated-android-branch -> automated-android-branch
```

.github/workflows/ci.yml
complexspaces: This is just a testing artifact, yes. I didn't want to waste CI resources when I was pushing to this branch a lot during development. I'll make sure to undo this before merging.

.github/workflows/android_builds.yml
cpu: > The downside of it seems to be that it treats our pushes to working branches in the repository as rule bypasses as well, which causes extra noise locally

Interesting! Thanks for testing.

I'm in the habit of pushing branches to a fork and opening PRs from there, so I don't think the noise would bother me but I can see how its annoying for others that don't want the overhead of managing branches on a fork.

complexspaces: To clarify, I am personally fine with the noise since I know what it means, and the gained benefit of locking this job down tightly feels very worth it. If I ever get tired of it I could also switch to using a fork. IMO given that we should keep the rule before merging this and live with the very minor downsides.

complexspaces: We could also consider adding a CODEOWNERS rule to this file to require another approval to prevent accidentally (though not malicious) updates to this workflow that are incorrect and have security consequences.

cpu: > We could also consider adding a CODEOWNERS rule to this file to require another approval

I wouldn't mind that extra caution if you wouldn't :-)

complexspaces:
Closing in favor of #251

PR #35, opened 2023-09-27
Error less often when native certificates fail to load on UNIX

complexspaces:
In #24 when `webpki-roots` was refactored out, https://github.com/rustls/rustls-platform-verifier/commit/5351caaf0f9bf94fb7c3fdea4ffd9ac86dc24c93 changed how errors from `rustls-native-certs` were handled. Prior to the change, we logged any errors and continued onward. But today we return an error and fail out of verifier initialization when certificates from the system's CA bundle fail to parse. Combined with issues such as https://github.com/rustls/rustls-native-certs/issues/48, this can lead to more failure cases.

This PR proposes changing the behavior to do two things differently:
- Load extra CA certificates before any from the system (will this be problematic? Which will take priority if they have the same key ID?)
- If any extra certificates have been loaded, log an error like before. If no extra roots are available, return an error since no verification will be possible because the combined root store will be empty.

### DIFF
--- a/src/verification/others.rs
+++ b/src/verification/others.rs
@@ -77,6 +77,17 @@ impl Verifier {
             }
         }
 
+        // Safety: There's no way for the mutex to be locked multiple times, so this is
+        // an infallible operation.
+        let mut extra_roots = self.extra_roots.try_lock().unwrap();
+        if !extra_roots.is_empty() {
+            let count = extra_roots.len();
+            root_store.add_trust_anchors(&mut extra_roots.drain(..));
+            log::debug!(
+                "Loaded {count} extra CA certificates in addition to possible system roots",
+            );
+        }
+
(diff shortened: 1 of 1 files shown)


cpu:
> (will this be problematic? Which will take priority if they have the same key ID?)

Neither Rustls nor Webpki will do any de-duplication of trust anchors (by subject, or by subject public key information). The `TrustAnchors` are tried in the order they're added to the `RootCertStore`, so the extra CA certificates would be tried before the system provided copies. We don't presently parse or consider authority or subject key identifiers. 

I don't believe this would cause any issues. Each CA certificate's subject and subject public key information will be processed independently until a valid path is found, budget for path building is expended, or all candidate CA certificates have been considered.

cpu (approved):
This looks good to me. Sorry about introducing the regression!

Issue #37, opened 2023-10-11
Fix MSRV build err from `regex-automata` via `cargo-ndk`.


cpu:
The scheduled CI job is hitting a failure w/ the MSRV tasks: https://github.com/rustls/rustls-platform-verifier/actions/runs/6486318645

```
  Downloaded libc v0.2.149
error: failed to compile `cargo-ndk v2.12.7`, intermediate artifacts can be found at `/tmp/cargo-installX3AYQ8`

Caused by:
  package `regex-automata v0.4.1` cannot be built because it requires rustc 1.65 or newer, while the currently active rustc version is 1.64.0
Error: Process completed with exit code 101.
```

Putting this in the tracker so it doesn't get lost in the shuffle.

PR #38, opened 2023-10-13
ci: use --locked for MSRV cargo-ndk install

cpu:
A transitive dependency of `cargo-ndk`, `regex-automata`,increased its MSRV and so installing `cargo-ndk` at version 2.12.7 breaks in the MSRV CI task unless we use the pinned `Cargo.lock` version, or upgrade our MSRV in turn.

See https://github.com/rustls/rustls-platform-verifier/pull/30 for previous struggles related to `cargo-ndk` and maintaining a MSRV of 1.64.

Resolves https://github.com/rustls/rustls-platform-verifier/issues/37.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -78,7 +78,7 @@ jobs:
 
       - name: Install cargo-ndk.
         run: |
-          cargo install cargo-ndk --version 2.12.7
+          cargo install cargo-ndk --locked --version 2.12.7
           rustup target add aarch64-linux-android
 
       - name: Clippy (${{ matrix.os }})


djc (approved):

complexspaces (approved):
IIRC using `--locked` for CI is best practice anyway, so this looks good 👍

complexspaces:
Am I the only one seeing the "Merge without waiting for requirements to be met" box in GitHub's UI? This looks like a previously unknown side effect of the branch protection experiments in #33. I guess that needs revisited, again.

cpu:
> Am I the only one seeing the "Merge without waiting for requirements to be met" box in GitHub's UI? 

I can confirm I'm seeing the same thing. 

> This looks like a previously unknown side effect of the branch protection experiments in https://github.com/rustls/rustls-platform-verifier/pull/33. I guess that needs revisited, again.

Ah bummer. I can take a closer look ~Monday. In the mean time I'll just break-glass merge to fix the CI job.

Issue #39, opened 2023-11-02
Rewrite certificate helpers in Rust


cpu:
Presently the [mock verification test cases](https://github.com/rustls/rustls-platform-verifier/tree/main/src/tests/verification_mock) are generated using Go, and the [real world verification test cases](https://github.com/rustls/rustls-platform-verifier/tree/main/src/tests/verification_real_world) are updated using a `bash` script invoking `openssl`.

It's likely possible to rewrite both of these test helpers in Rust, making maintenance a little bit easier. The replacement for the `bash` tooling for the real world verification tests could also fetch intermediate certificates in addition to end entity certs.

Putting this issue in the tracker for a rainy day.

yerke:
Updated links to [mock verification test cases](https://github.com/rustls/rustls-platform-verifier/tree/main/rustls-platform-verifier/src/tests/verification_mock) and [real world verification test cases](https://github.com/rustls/rustls-platform-verifier/tree/main/rustls-platform-verifier/src/tests/verification_real_world).

cpu:
@yerke Thanks! 

I updated this issue to reflect that the real world verification helper script was written in #44. It's just the mock verification test cases that need similar treatment now.

The updated issue description uses permalinks and also provides a link to similar work I did in the upstream `rustls` repo for inspiration. I'm not planning to tackle this myself anytime soon but hopefully these pointers whoever decides to take a crack at it.

PR #40, opened 2023-11-02
Only check end-entity certificate revocation status on Android

cpu:
## Description

This branch proposes an alternative to https://github.com/rustls/rustls-platform-verifier/pull/34, addressing https://github.com/rustls/rustls-platform-verifier/issues/13 based on discussion with @complexspaces and others.

In summary, given the constraints[^1] imposed by the Android platform we feel that checking revocation information for only the end-entity certificate of a given validated certificate chain is the least-worst option. This is a closer match to how `Conscrypt` handles revocation checking (with the caveat that they also prevent fallback to CRLs while we, at least for now, allow this). In general this approach is easier to understand, and easier to align with other platform verifiers. 

This branch updates the Android `CertificateVerifier.kt` platform specific verifier logic used by the Rust code to configure revocation checking such that we:

* Only check revocation status for the end entity certificate
* Prefer OCSP, if stapled or otherwise
* Allow fallback to CRLs
* Allow soft failure for network issues

As a result of these changes the logic to determine if a trust anchor was one installed in the system trust store is no longer required, so it falls away along with its dedicated unit test. A new real world verification test case is added for a Let's Encrypt certificate chain that previously failed due to the R3 intermediate in use not providing an AIA OCSP URI. To make keeping the shorted lived EE certificate up-to-date the bash script for fetching valid example certs is generalized to include this test case as well. In the future (https://github.com/rustls/rustls-platform-verifier/issues/39) it might be nice to rewrite this in Rust.

[^1]: We can't control revocation checking preferences per-certificate, it must be configured for the whole chain. If we check revocation status for the whole chain preferring OCSP, and a certificate is missing AIA OCSP URI information, it is hard failure. We can't easily check ahead of time whether a given certificate has an AIA OCSP URI. Stapled responses are only available for EE certs. Network access to fetch revocation information can't be controlled (e.g. to only use cached/local information).


complexspaces (approved):
I apologize for taking so long to approve this. The code is perfectly fine but I've been getting really bugged about documentation around revocation for the platform verifier, and that threw me off.

In hindsight we should have merged this two weeks ago and followed up with documentation later 😅

cpu:
@complexspaces Thanks for the review! I understand this is a tricky set of compromises to navigate and taking time to consider carefully them is very reasonable.

> In hindsight we should have merged this two weeks ago and followed up with documentation later 😅

Is there any way I can help with this as follow-up? If you have super rough bullet-points about what you'd like covered I could take a crack at drafting some text if it would be helpful. LMK.

PR #41, opened 2023-11-16
tests: update agilebits_com_valid_1.crt

cpu:
Scheduled CI jobs are currently failing the real world test suite ([example](https://github.com/rustls/rustls-platform-verifier/actions/runs/6894728399/job/18757180707)).

The root cause is the `agilebits.com` real world test certificate expired yesterday, on Nov 15 2023. This branch replaces it with an updated copy fetched from `agilebits.com` that expires Oct 16 2024.  Additionally, the chain of intermediates no
longer matches the ones used by `1password.com`, so we add the required intermediates and remove the "SHARED_CHAIN" bits from the realworld verification tests.


cpu:
Looks like there's some more adjustment needed:
```
---- tests::verification_real_world::tests::unrelated_chain_valid_for_unrelated_domain stdout ----
thread 'tests::verification_real_world::tests::unrelated_chain_valid_for_unrelated_domain' panicked at src/tests/mod.rs:46:9:
assertion `left == right` failed
  left: Err(InvalidCertificate(UnknownIssuer))
 right: Ok(())
```

Previously this test case was sharing its intermediate chain w/ the `VALID_1PASSWORD_COM_CHAIN` testcase, but it looks like the `agilebits.com` chain has diverged. Its issuer has serial `07:73:12:38:0b:9d:66:88:a3:3b:1e:d9:bf:9c:cd:a6:8e:0e:0f` while the first intermediate in the shared chain has serial `07:73:12:4a:4b:cb:d4:4e:c7:b5:3b:ea:f1:94:84:2d:3a:0f:a1`. I'll push another revision.

djc (approved):

complexspaces (approved):

PR #42, opened 2023-12-05
Cargo: update to rustls 0.22, associated updates

cpu:
## Description

This branch updates to Rustls 0.22 and brings in the associated dependency updates this requires. For the time being, this branch continues to unconditionally use *ring* as the crypto provider. Follow-up work to expose this as a choice (e.g allowing aws-lc-rs as a provider) may be interesting.

### Deps:

* updated rustls 0.21 -> 0.22.1
* added pki-types 1.0

Linux deps:
* rustls-native-certs 0.6 -> 0.7
* webpki 0.101 -> 0.102

Android deps:
* webpki 0.101 -> 0.102

WASM32 deps:
* webpki-roots 0.25 -> 0.26

### Summary of breaking change updates:

* We use rustls 0.22.1 in specific to benefit from the `pki_types` re-export, removing the need to add that as our own dep with matching version.
* `ServerName`, `Certificate`, and `OwnedTrustAnchor` types are now sourced from `pki_types`, with an associated generic lifetime. The `OwnedTrustAnchor` type is now just `TrustAnchor`.
* The 'dangerous' rustls crate feature was removed, and associated items moved into new locations with the import path emphasizing danger.
* "Other error" types changed to use a specific `rustls::OtherError` inner variant.
* `SystemTime` for verifiers replaced with `pki_types::UnixTime`.
* Default fns on `ServerCertVerifier` trait were removed, must be reconstituted with `rustls::verify_tls12_signature`, `rustls::verify_tls13_signature` and `WebPkiSupportedAlgorithms.supported_schemes` using a `CryptoProvider`.
* `ServerName` now supports a `to_str` operation, avoiding the need to `match` and handle unsupported name types.
* `WebPkiVerifier` was renamed to `WebPkiServerVerifier`, handled as an `Arc` and constructed with a builder.

Reviewers will likely be most interested in some of the changes associated with the new `pki_types::UnixTime` type. Since we no longer receive a `SystemTime` from Rustls some of the code adjusting the verification timestamp into platform specific representations needed to change. I think I've handled this correctly but would appreciate more detailed review on this aspect of the branch.

### DIFF
--- a/rustls-platform-verifier/src/tests/verification_mock/mod.rs
+++ b/rustls-platform-verifier/src/tests/verification_mock/mod.rs
@@ -18,7 +18,9 @@
 use super::TestCase;
 use crate::tests::{assert_cert_error_eq, verification_time};
 use crate::verification::{EkuError, Verifier};
-use rustls::{client::ServerCertVerifier, CertificateError, Error as TlsError};
+use rustls::client::danger::ServerCertVerifier;
+use rustls::pki_types;
+use rustls::{CertificateError, Error as TlsError, OtherError};
 use std::convert::TryFrom;
 use std::net::IpAddr;
 use std::sync::Arc;
@@ -77,18 +79,28 @@ const LOCALHOST_IPV6: &str = "::1";
 #[cfg(any(test, feature = "ffi-testing"))]
 #[cfg_attr(feature = "ffi-testing", allow(dead_code))]
 pub(super) fn verification_without_mock_root() {
+    // Since Rustls 0.22 constructing a webpki verifier (like the one backing Verifier on unix
+    // systems) without any roots produces `OtherError(NoRootAnchors)` - since our FreeBSD CI
+    // runner fails to find any roots with openssl-probe we need to provide webpki-roots here
+    // or the test will fail with the `OtherError` instead of the expected `CertificateError`.
+    #[cfg(target_os = "freebsd")]
+    let verifier = Verifier::new_with_extra_roots(webpki_roots::TLS_SERVER_ROOTS.iter().cloned());
+
+    #[cfg(not(target_os = "freebsd"))]
     let verifier = Verifier::new();
 
-    let server_name = rustls::client::ServerName::try_from(EXAMPLE_COM).unwrap();
-    let end_entity = rustls::Certificate(ROOT1_INT1_EXAMPLE_COM_GOOD.to_vec());
-    let intermediates = [rustls::Certificate(ROOT1_INT1.to_vec())];
+    let server_name = pki_types::ServerName::try_from(EXAMPLE_COM).unwrap();
+    let end_entity = pki_types::CertificateDer::from(ROOT1_INT1_EXAMPLE_COM_GOOD);
+    let intermediates = [pki_types::CertificateDer::from(ROOT1_INT1)];
 
     // Fails because the server cert has no trust root in Windows, and can't since it uses a self-signed CA.
+    // Similarly on UNIX platforms using the Webpki verifier, it can't fetch extra certificates through
+    // AIA chasing or other mechanisms, and so we know this test will correctly verify an unknown
+    // root in a chain fails validation.
     let result = verifier.verify_server_cert(
         &end_entity,
         &intermediates,
         &server_name,
-        &mut std::iter::empty(),
         &[],
         verification_time(),
     );
@@ -247,7 +259,7 @@ mock_root_test_cases! {
         stapled_ocsp: None,
         verification_time: verification_time(),
         expected_result: Err(TlsError::InvalidCertificate(
-            CertificateError::Other(Arc::from(EkuError)))),
+            CertificateError::Other(OtherError(Arc::from(EkuError))))),
         other_error: Some(EkuError),
     },
     wrong_eku_ipv4 [ any(windows, unix) ] => TestCase {
@@ -256,7 +268,7 @@ mock_root_test_cases! {
         stapled_ocsp: None,
         verification_time: verification_time(),
         expected_result: Err(TlsError::InvalidCertificate(
-            CertificateError::Other(Arc::from(EkuError)))),
+            CertificateError::Other(OtherError(Arc::from(EkuError))))),
         other_error: Some(EkuError),
     },
     wrong_eku_ipv6 [ any(windows, unix) ] => TestCase {
@@ -265,7 +277,7 @@ mock_root_test_cases! {
         stapled_ocsp: None,
         verification_time: verification_time(),
         expected_result: Err(TlsError::InvalidCertificate(
-            CertificateError::Other(Arc::from(EkuError)))),
+            CertificateError::Other(OtherError(Arc::from(EkuError))))),
         other_error: Some(EkuError),
     },
 }
@@ -277,30 +289,23 @@ fn test_with_mock_root<E: std::error::Error + PartialEq + 'static>(test_case: &T
     let mut chain = test_case
         .chain
         .iter()
-        .map(|bytes| rustls::Certificate(bytes.to_vec()));
+        .map(|bytes| pki_types::CertificateDer::from(*bytes));
 
     let end_entity = chain.next().unwrap();
-    let intermediates: Vec<rustls::Certificate> = chain.collect();
+    let intermediates: Vec<pki_types::CertificateDer<'_>> = chain.collect();
 
-    let server_name = rustls::client::ServerName::try_from(test_case.reference_id).unwrap();
+    let server_name = pki_types::ServerName::try_from(test_case.reference_id).unwrap();
 
     if test_case.reference_id.parse::<IpAddr>().is_ok() {
-        assert!(matches!(
-            server_name,
-            rustls::client::ServerName::IpAddress(_)
-        ));
+        assert!(matches!(server_name, pki_types::ServerName::IpAddress(_)));
     } else {
-        assert!(matches!(
-            server_name,
-            rustls::client::ServerName::DnsName(_)
-        ));
+        assert!(matches!(server_name, pki_types::ServerName::DnsName(_)));
     }
 
     let result = verifier.verify_server_cert(
         &end_entity,
         &intermediates,
         &server_name,
-        &mut std::iter::empty(),
         test_case.stapled_ocsp.unwrap_or(&[]),
         test_case.verification_time,
     );
--- a/rustls-platform-verifier/src/tests/verification_real_world/mod.rs
+++ b/rustls-platform-verifier/src/tests/verification_real_world/mod.rs
@@ -37,7 +37,9 @@
 use super::TestCase;
 use crate::tests::{assert_cert_error_eq, verification_time};
 use crate::Verifier;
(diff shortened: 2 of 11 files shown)


cpu:
> cpu marked this pull request as ready for review now

The main blocker that had this in draft status was the Reqwest compatibility unit test, but I think we should land https://github.com/rustls/rustls-platform-verifier/pull/43 to address that. I've marked this branch ready for review and updated the PR description to remove the TODO notes.

src/verification/android.rs
```diff
@@ -35,7 +38,7 @@ enum VerifierStatus {
 const AUTH_TYPE: &str = "RSA";
 
 /// A TLS certificate verifier that utilizes the Android platform verifier.
-#[derive(Default)]
+#[derive(Default, Debug)]
```
djc, line 41: Tiny nit: alphabetic?

(There is at least one other instance.)

djc (approved):
LGTM!

Cargo.toml
```diff
@@ -37,26 +37,27 @@ cert-logging = ["base64"]
 docsrs = ["jni", "once_cell"]
 
 [dependencies]
-rustls = { version = "0.21", features = ["dangerous_configuration", "tls12", "logging"] }
+rustls = { version = "0.22", features = ["tls12", "logging"] }
```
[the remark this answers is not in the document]
cpu: Good idea, thanks. Implemented.

src/verification/android.rs
cpu: Fixed throughout :+1:

cpu:
> cpu marked this pull request as draft now

Flipping this to draft - needs a rebase :-)

cpu:
> Flipping this to draft - needs a rebase :-)

Rebased and ready to go.

cpu:
> cpu force-pushed the cpu-rustls-0.22 branch from 239f0b3 to a3d1842 

Rebased again to pick up the changes from https://github.com/rustls/rustls-platform-verifier/pull/50

cpu:
> cpu force-pushed the cpu-rustls-0.22 branch from a3d1842 to 1ab5d96 

rebased again for https://github.com/rustls/rustls-platform-verifier/pull/55

cpu:
> I'll fix shortly.

Done. 

Besides fixing the webpki-roots version I had to tweak the `verification_without_mock_root` test to use the same strategy we used in #55: initializing the `Verifier` under test with `Verifier::new_with_extra_roots` for FreeBSD. 

As we know already, in FreeBSD CI the `openssl-probe` wasn't populating the `Verifier` with roots on its own. On Unix-likes w/ Rustls 0.21 the test's `Verifier::new()` would construct a webpki verifier under the hood that had no roots configured, and we get the expected `CertificateError::UnknownIssuer` error the test wants. With Rustls 0.22 this test fails because the webpki verifier construction was changed to error when initialized with an empty root store, resulting in a different err than the test expects. Switching the `Verifier` constructor out for `Verifier::new_with_extra_roots` results in the webpki verifier constructor succeeding, and then the test fails in the expected way because wepki-roots doesn't contain the mock root trust anchor.

rustls-platform-verifier/src/verification/android.rs
```diff
@@ -305,4 +284,38 @@ impl ServerCertVerifier for Verifier {
...
+        &self,
+        message: &[u8],
+        cert: &pki_types::CertificateDer<'_>,
+        dss: &DigitallySignedStruct,
+    ) -> Result<HandshakeSignatureValid, TlsError> {
+        verify_tls12_signature(
+            message,
+            cert,
+            dss,
+            &rustls::crypto::ring::default_provider().signature_verification_algorithms,
```
complexspaces, line 298: Would it be better to store the provider on the verifier structure so it can be referenced through `&self`? Otherwise this is going to re-allocate two vecs on each signature (`cipher_suites` and `kx_groups`).

This comment also applies to the other platform verifier implementations.

rustls-platform-verifier/src/verification/android.rs
```diff
@@ -305,4 +284,38 @@ impl ServerCertVerifier for Verifier {
...
+    ) -> Result<HandshakeSignatureValid, TlsError> {
+        verify_tls13_signature(
+            message,
+            cert,
+            dss,
+            &rustls::crypto::ring::default_provider().signature_verification_algorithms,
+        )
+    }
+
+    fn supported_verify_schemes(&self) -> Vec<SignatureScheme> {
```
complexspaces, line 316: Out of curiosity, howcome supported schemes uses a `Vec` instead of `&'static []`? Is it expected these could change at runtime somehow inside a provider?

rustls-platform-verifier/src/verification/apple.rs
```diff
@@ -183,36 +183,26 @@ impl Verifier {
 impl ServerCertVerifier for Verifier {
     fn verify_server_cert(
         &self,
-        end_entity: &rustls::Certificate,
-        intermediates: &[rustls::Certificate],
-        server_name: &rustls::ServerName,
-        _scts: &mut dyn Iterator<Item = &[u8]>,
```
complexspaces, line 189: This is somewhat of a downside since Apple [does provide a function for this](https://developer.apple.com/documentation/security/3081274-sectrustsetsignedcertificatetime?language=objc), but then again Apple is the only platform that supports them.

rustls-platform-verifier/src/verification/others.rs
```diff
@@ -62,18 +68,23 @@ impl Verifier {
...
         let mut root_store = rustls::RootCertStore::empty();
 
         // For testing only: load fake root cert, instead of native/WebPKI roots
         #[cfg(any(test, feature = "ffi-testing", feature = "dbg"))]
         {
             if let Some(test_root) = &self.test_only_root_ca_override {
-                let (added, ignored) = root_store.add_parsable_certificates(&[test_root.clone()]);
+                let (added, ignored) =
+                    root_store.add_parsable_certificates([pki_types::CertificateDer::from(
+                        test_root.clone(),
```
complexspaces, line 80: Since the `test_roots` are `&'static`, this clone may not be needed?

rustls-platform-verifier/src/verification/others.rs
```diff
@@ -62,18 +68,23 @@ impl Verifier {
...
+                    )]);
                 if (added != 1) || (ignored != 0) {
                     panic!("Failed to insert fake, test-only root trust anchor");
                 }
-                return Ok(WebPkiVerifier::new(root_store, None));
+                return WebPkiServerVerifier::builder(root_store.into())
+                    .build()
+                    .map_err(|e| TlsError::Other(OtherError(Arc::new(e))));
```
complexspaces, line 87: Nit: Given this is a test and we're already panicking above if no roots were added, should this just `.unwrap()`?

complexspaces (changes requested):
This looks good for the most part, I just have some nits/questions regarding the new types and crypto backends.

rustls-platform-verifier/src/verification/android.rs
cpu: I'm not sure. The return type for this fn has remained unchanged since it was introduced ~4 years ago: https://github.com/rustls/rustls/commit/1e98fa2eb1004b0ddbf68bdbcbf960891667e10b

rustls-platform-verifier/src/verification/apple.rs
cpu: I think this would have to be addressed upstream in Rustls.

rustls-platform-verifier/src/tests/verification_mock/mod.rs
```diff
@@ -277,30 +286,23 @@ fn test_with_mock_root<E: std::error::Error + PartialEq + 'static>(test_case: &T
     let mut chain = test_case
         .chain
         .iter()
-        .map(|bytes| rustls::Certificate(bytes.to_vec()));
+        .map(|bytes| pki_types::CertificateDer::from(bytes.to_vec()));
```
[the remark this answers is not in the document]
cpu: Fixed.

rustls-platform-verifier/src/tests/verification_real_world/mod.rs
```diff
@@ -121,31 +123,20 @@ fn real_world_test<E: std::error::Error>(test_case: &TestCase<E>) {
...
+    let verifier = Verifier::new_with_extra_roots(webpki_roots::TLS_SERVER_ROOTS.iter().cloned());
 
     #[cfg(not(target_os = "freebsd"))]
     let verifier = Verifier::new();
 
     let mut chain = test_case
         .chain
         .iter()
-        .map(|bytes| rustls::Certificate(bytes.to_vec()));
+        .map(|bytes| pki_types::CertificateDer::from(bytes.to_vec()));
```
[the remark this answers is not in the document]
cpu: Fixed.

rustls-platform-verifier/src/verification/android.rs
cpu: Fixed.

rustls-platform-verifier/src/verification/others.rs
cpu: Fixed.

rustls-platform-verifier/src/verification/others.rs
```diff
@@ -169,15 +168,52 @@ impl ServerCertVerifier for Verifier {
...
+    }
+
+    fn supported_verify_schemes(&self) -> Vec<SignatureScheme> {
+        let verifier = match self.inner.get_or_try_init(|| self.init_verifier()) {
+            Ok(v) => v,
+            Err(_) => return Vec::default(),
+        };
+        verifier.supported_verify_schemes()
+    }
+}
```
cpu, line 201: This trait fn doesn't support returning a `Result` so if the once cell verifier init fails it returns an empty vec. Alternatively we could maintain a `CryptoProvider` in the other verifier struct to use instead of the once-cell wrapped webpki verifier. I thought this approach was better in sum.

rustls-platform-verifier/src/verification/android.rs
complexspaces: Thanks for the extra context, I didn't realize this wasn't a new introduction with `0.22`. I think that makes sense for the most part, so I'll leave it be 👍

rustls-platform-verifier/src/verification/others.rs
complexspaces: I think this approach seems reasonable to me. Now that we have 4 places calling `get_or_try_init`, I wonder if a helper function on `&self`, something like `get_or_init_verifier(), would be useful to abstract over the `OnceCell` boilerplate (`get_or_try_init(|| self.init_verifier()))` more.

complexspaces (approved):
This looks great now, thanks!

There's just the one small thread left about maybe abstracting some `OnceCell` boilerplate away, but its not a blocker.

rustls-platform-verifier/src/verification/others.rs
cpu: Sounds reasonable, I can tack that on

cpu: Fixed.

complexspaces:
The latest cleanup changes still look good to me, so LGTM.

PR #43, opened 2023-12-14
remove reqwest unit tests, dev-dependency

cpu:
Previously the `verification` mod had a unit test, `can_verify_server_cert`, that ensured using the platform verifier with `reqwest` could be done without error.

The reason for this was that the `reqwest` API consumes custom verifiers to use with a Rustls client config as `&dyn Any` inputs, and then downcasts at runtime to the required `Arc<dyn ServerCertVerifier>` - this means that if `rustls-platform-verifier` uses a different Rustls version than `reqwest` a runtime panic would occur.

However, having this unit test in place means we can't update `rustls-platform-verifier` to a new Rustls release until the `reqwest` ecosystem first updates. This is suboptimal, as `reqwest` itself has many dependencies that need similar updates.

This commit removes the unit test. Ensuring the Rustls versions match should be handled by downstream consumers that have chosen to use `reqwest`. There are other libraries one might use `rustls-platform-verifier` with, and we shouldn't block useful updates to this crate on `reqwest`. In general one already has to be careful about mixing/matching Rustls versions across dependencies, the fact that `reqwest` makes this a runtime concern is unfortunate, but not a great reason to avoid keeping this crate in sync with the rest of the 1st-party Rustls ecosystem.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,21 +3,6 @@
 version = 3
 
 [[package]]
-name = "addr2line"
-version = "0.21.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "8a30b2e23b9e17a9f90641c7ab1549cd9b44f296d3ccbf309d2863cfe398a0cb"
-dependencies = [
- "gimli",
-]
-
-[[package]]
-name = "adler"
-version = "1.0.2"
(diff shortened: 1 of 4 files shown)


djc (approved):
It's not great but I guess it's an improvement.

complexspaces (approved):
Thanks!

PR #46, opened 2023-12-29
Fix Rust 1.75 clippy finding, expired real world test certificate

cpu:
* The Let's Encrypt test certificate expired, breaking the real world verification test suite.
* Rust 1.75 is flagging the `pub use tests::ffi::*;` in `src/lib.rs` that's behind the `ffi-testing` feature flag as unused. This branch allows that unused import.


cpu:
> The Let's Encrypt test certificate expired, breaking the real world verification test suite.

Since this is going to keep happening every ~90d (_who thought automated certificate issuance was a good idea? Sounds crazy to me! :zany_face:_) we should probably fix this more robustly. Some ideas:

1. We could use something like https://github.com/rustls/rustls-platform-verifier/pull/47 to keep the test cert up-to-date (but then have to contend with CI permissions).
2. We could switch to a "live" test case that depends on a remote server we assume is correctly rotating the cert every ~60d instead of using a vendored copy. This is likely to flake in different ways - Rustls has a similar test suite and moved it to a separate daily CI task for that reason.
3. We remove the LE test case deciding its more trouble than its worth.
4. Status quo: we let CI fail and a human can use our shiny new Rust script to update the test files.

ctz:
Would running the tests under `faketime` work?

cpu:
> edit: and, by implication, only on linux.

For another repo that would probably be an OK trade-off but I think for this one the platform coverage is important :-(

complexspaces:
I don't think that we need any platform-specific time-mocking. There's just a current defficency in both the mock and real world tests where we pass in `SystemTime::now()` into the platform verifier instead of a hardcoded datetime. You can see that [here](https://github.com/rustls/rustls-platform-verifier/blob/main/src/tests/verification_real_world/mod.rs#L148) and [here](https://github.com/rustls/rustls-platform-verifier/blob/main/src/tests/verification_mock/mod.rs#L292). If we update these to use a hardcoded date, such as today for example, the certificates should last much longer and only be prone to root expiry.

cpu:
> If we update these to use a hardcoded date, such as today for example, the certificates should last much longer and only be prone to root expiry. I apologize for not calling that out earlier, I only remembered it when thinking about a better solution then something Linux-specific.


Ah! That does sound best. Can we merge this to fix CI and I will follow up on that approach?

complexspaces (approved):

complexspaces:
That sounds good to me. I agree its better to get CI unbroken first.

PR #47, opened 2023-12-29
ci: add cert-update cron workflow

cpu:
This branch adds a CI workflow that runs once a week to create a PR updating the real world verification test certificates (if required).

This will help prevent CI breaking when the vendored Let's Encrypt exemplar certificate expires (see https://github.com/rustls/rustls-platform-verifier/pull/46).

Like https://github.com/rustls/rustls-platform-verifier/pull/33 the main consideration here is whether we're comfortable with the `contents: write` permission on the `test-cert-update` job to allow it to push branches to open pull requests from. I've opened this as a draft separate from #46 while we think on that aspect.


cpu:
Better solved with a fixed verification time.

PR #48, opened 2024-01-03
Remove crate-level test verifier constructor

complexspaces:
This PR makes a few small cleanup changes before the 0.1 release, which only affect test code. Since the `Verifier` struct became part of the public API, the `verifier_for_testing` function has been redundant and more confusing then helpful.


complexspaces:
As this only affects test-only, `pub(crate)` code, I'm going to merge it without a review as to not need an extra rebase for the 0.1 release prep.

cpu (commented):
Nice tidy :-)

PR #49, opened 2024-01-03
Prepare 0.1

complexspaces:
Proposed initial CHANGELOG entry:
- Initial release with support for Android, iOS, macOS, Windows, and Linux/UNIX-like platforms. See the README for details.

These are the steps remaining for the release after this merges:
- [ ] Generate the Android Maven artifacts
- [ ] Create Git tag (TODO: Should we check-in the generated artifacts for the tag to crates.io users have another comparison reference?)
- [ ] Push tag
- [ ] `cargo publish` (`--allow-dirty` may be needed based on the above)
- [ ] Create companion GitHub release

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,0 +1,1 @@
+The detailed list of changes in each release can be found at https://github.com/rustls/rustls-platform-verifier/releases.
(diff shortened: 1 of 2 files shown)


cpu (approved):
:rocket: :rocket: :rocket: :rocket: :rocket:

complexspaces:
This release is now finished. The associated GitHub release can be found [here](https://github.com/rustls/rustls-platform-verifier/releases) and has the pre-compiled Android artifacts attached in a reproducible format as well. It was decided that a Git tag with its own commit had poor enough UX in Git's CLI that it wasn't worth putting the artifacts there.

PR #50, opened 2024-01-03
tests: use a fixed SystemTime for certificate validation

cpu:
Fixing the `SystemTime` that we pass to the certificate verifier for the real world and mock verification tests will ensure that the tests don't start to fail just because the vendored certificates have expired.

[Suggested by @complexspaces](https://github.com/rustls/rustls-platform-verifier/pull/46#issuecomment-1872275559) in https://github.com/rustls/rustls-platform-verifier/pull/46

### DIFF
--- a/rustls-platform-verifier/src/tests/verification_mock/mod.rs
+++ b/rustls-platform-verifier/src/tests/verification_mock/mod.rs
@@ -21,7 +21,7 @@
 ))]
 
 use super::TestCase;
-use crate::tests::assert_cert_error_eq;
+use crate::tests::{assert_cert_error_eq, verification_time};
 use crate::verification::{EkuError, Verifier};
 use rustls::{client::ServerCertVerifier, CertificateError, Error as TlsError};
 use std::convert::TryFrom;
@@ -95,7 +95,7 @@ pub(super) fn verification_without_mock_root() {
         &server_name,
         &mut std::iter::empty(),
         &[],
-        std::time::SystemTime::now(),
+        verification_time(),
     );
 
     assert_eq!(
@@ -120,6 +120,7 @@ mock_root_test_cases! {
         reference_id: EXAMPLE_COM,
         chain: &[ROOT1_INT1_EXAMPLE_COM_GOOD, ROOT1_INT1],
         stapled_ocsp: None,
+        verification_time: verification_time(),
         expected_result: Ok(()),
         other_error: no_error!(),
     },
@@ -127,6 +128,7 @@ mock_root_test_cases! {
         reference_id: LOCALHOST_IPV4,
         chain: &[ROOT1_INT1_LOCALHOST_IPV4_GOOD, ROOT1_INT1],
         stapled_ocsp: None,
+        verification_time: verification_time(),
         expected_result: Ok(()),
         other_error: no_error!(),
     },
@@ -134,6 +136,7 @@ mock_root_test_cases! {
         reference_id: LOCALHOST_IPV6,
         chain: &[ROOT1_INT1_LOCALHOST_IPV6_GOOD, ROOT1_INT1],
         stapled_ocsp: None,
+        verification_time: verification_time(),
         expected_result: Ok(()),
         other_error: no_error!(),
     },
@@ -141,6 +144,7 @@ mock_root_test_cases! {
         reference_id: EXAMPLE_COM,
         chain: &[ROOT1_INT1_EXAMPLE_COM_GOOD, ROOT1_INT1],
         stapled_ocsp: Some(include_bytes!("root1-int1-ee_example.com-good.ocsp")),
+        verification_time: verification_time(),
         expected_result: Ok(()),
         other_error: no_error!(),
     },
@@ -148,6 +152,7 @@ mock_root_test_cases! {
         reference_id: LOCALHOST_IPV4,
         chain: &[ROOT1_INT1_LOCALHOST_IPV4_GOOD, ROOT1_INT1],
         stapled_ocsp: Some(include_bytes!("root1-int1-ee_127.0.0.1-good.ocsp")),
+        verification_time: verification_time(),
         expected_result: Ok(()),
         other_error: no_error!(),
     },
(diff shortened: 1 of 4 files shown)


rustls-platform-verifier/src/lib.rs
```diff
@@ -71,3 +72,12 @@ pub fn tls_config() -> ClientConfig {
...
     Arc::new(Verifier::new_with_fake_root(root))
 }
+
+/// Return a fixed [SystemTime] for certificate validation purposes.
+///
+/// We fix the "now" value used for certificate validation to a fixed point in time at which
+/// we know the test certificates are valid. This must be updated if the test certificates
+/// are regenerated.
+pub fn verification_time() -> SystemTime {
+    SystemTime::UNIX_EPOCH + Duration::from_secs(1_704_304_988)
```
cpu, line 82: Unix epoch at the time of writing this PR.

Will need to adjust after https://github.com/rustls/rustls-platform-verifier/pull/42 lands to avoid `SystemTime`.

complexspaces, line 82: Can we leave a comment saying which datetime this refers to for easy reading/review?

rustls-platform-verifier/src/lib.rs
```diff
@@ -71,3 +72,12 @@ pub fn tls_config() -> ClientConfig {
 pub fn verifier_for_dbg(root: &[u8]) -> Arc<dyn rustls::client::ServerCertVerifier> {
     Arc::new(Verifier::new_with_fake_root(root))
 }
+
+/// Return a fixed [SystemTime] for certificate validation purposes.
+///
+/// We fix the "now" value used for certificate validation to a fixed point in time at which
+/// we know the test certificates are valid. This must be updated if the test certificates
+/// are regenerated.
+pub fn verification_time() -> SystemTime {
```
complexspaces, line 81: This should probably be `pub(crate)` and `#[cfg(test)]` too.

complexspaces (changes requested):
There's two other items that should be fixed/acknowledged here as well:
- [This comment block](https://github.com/rustls/rustls-platform-verifier/blob/65b2a97aff062585d91c97ae3b7b1d17fbcd7b62/rustls-platform-verifier/src/tests/verification_real_world/mod.rs#L25-L30) should be deleted because its no longer accurate.
- We should see if the [OSCP stapling test](https://github.com/rustls/rustls-platform-verifier/blob/65b2a97aff062585d91c97ae3b7b1d17fbcd7b62/rustls-platform-verifier/src/tests/verification_real_world/mod.rs#L223) can be uncommented finally. If it doesn't work right away or needs medium effort to revive it can be punted to a followup issue.

rustls-platform-verifier/src/lib.rs
cpu: Oh! I actually meant to add this to `src/tests/mod.rs` but put it in `src/lib.rs` by mistake. Thanks for flagging. I'll move this and adjust visibility (I don't think the `cfg` gate is required if it's under `src/tests`)

cpu:
> We should see if the [OSCP stapling test](https://github.com/rustls/rustls-platform-verifier/blob/65b2a97aff062585d91c97ae3b7b1d17fbcd7b62/rustls-platform-verifier/src/tests/verification_real_world/mod.rs#L223) can be uncommented finally. If it doesn't work right away or needs medium effort to revive it can be punted to a followup issue.

let me see if I can get it working. I think at a minimum we'll need to refresh the vendored OCSP resp. The one in-repo expired in 2021:

```
$ openssl ocsp --respin rustls-platform-verifier/src/tests/verification_real_world/revoked_badssl_com_1.ocsp -noverify -text | grep "Next Update"
    Next Update: Dec  9 01:24:01 2021 GMT
```

cpu:
> The one in-repo expired in 2021:

Oh, well the `revoked.badssl.com` endpoint is also serving an expired cert (which I suppose makes sense, they aren't going to re-issue and re-revoke certs to keep the expiry up-to-date). I think we'll need to switch to using a fixed time per testcase for this.

rustls-platform-verifier/src/lib.rs
cpu: Done.

rustls-platform-verifier/src/lib.rs
cpu: Fixed.

cpu:
> This comment block should be deleted because its no longer accurate.

Fixed.

> We should see if the OSCP stapling test can be uncommented finally. If it doesn't work right away or needs medium effort to revive it can be punted to a followup issue.

I [started on this](https://github.com/rustls/rustls-platform-verifier/commit/c9954442e488c5df4d0b7a2f7bcbe25aabdead74), but surprisingly the Windows verifier seems to be [returning `Ok(())`](https://github.com/cpu/rustls-platform-verifier/actions/runs/7401734560/job/20138120366) when I would have expected it to honour the stapled OCSP response and return revoked. This will need more investigation so I've pushed the rest of the feedback in the meantime.

rustls-platform-verifier/src/tests/verification_mock/mod.rs
```diff
@@ -120,41 +120,47 @@ mock_root_test_cases! {
         reference_id: EXAMPLE_COM,
         chain: &[ROOT1_INT1_EXAMPLE_COM_GOOD, ROOT1_INT1],
         stapled_ocsp: None,
+        verification_time: verification_time(),
```
cpu, line 123: note: it looks a little silly to have a `verification_time` field on the `TestCase` struct, but then to populate it universally with `verification_time()` - this is done so that we can use a different date in the past for the stapled OCSP revocation test case. Enabling that is a WIP but I left this groundwork in place.

cpu:
> I started on this, but surprisingly the Windows verifier seems to be returning Ok(()) when I would have expected it to honour the stapled OCSP response and return revoked.

Conditionally adding `CERT_CHAIN_REVOCATION_CHECK_CACHE_ONLY` when we have an OCSP staple doesn't seem to do the trick either. I'm stumped (and tired of iterating in CI :sweat_smile:) so I'm going to pause on this for now. I'll file an issue for follow-up.

rustls-platform-verifier/src/tests/verification_mock/mod.rs
complexspaces: That seems reasonable to me. It seems plausible we might want to expand testing in the future to check prior-to validity dates, etc, so I think this is definitely worthwhile.

complexspaces (approved):
Thanks for jumping on that followup issue quick!

Issue #53, opened 2024-01-04
FreeBSD support


Ralith:
This is blocking Quinn from shipping rustls-platform-verifier helpers out of the box (https://github.com/quinn-rs/quinn/pull/1734). May be as simple as replacing `cfg(target_os = "linux")` with `cfg(all(unix, not(target_os = "macos"), not(target_os = "ios")))`.

cpu:
That seems reasonable. I think the more interesting aspect will be CI coverage. It [looks like](https://github.com/quinn-rs/quinn/blob/55234e178fdca81cd51a5bfb520cb912de14f72e/.github/workflows/rust.yml#L11-L32) Quinn is using `vmactions/freebsd-vm@v1` for that. I suppose we should follow suit?

complexspaces:
I agree this is just an oversight and FreeBSD support should be expected. `vmactions/freebsd-vm@v1` seems like it's pretty widely used on GitHub.

As for the proper gate, I think we want this for the `other.rs` module. Android has its own implementation, not just Apple's platforms:
```rust
cfg(all(unix, not(target_os = "android") not(target_os = "macos"), not(target_os = "ios")))
```

cpu:
https://github.com/rustls/rustls-platform-verifier/pull/55

PR #55, opened 2024-01-04
Support FreeBSD

cpu:
This branch relaxes the cfg gates that previously were Linux specific to allow Unix generally. Care is taken to ensure we still handle MacOS/iOS/Android specially where required.

FreeBSD in CI seems to be unable to use openssl-probe to find the system CA bundle, so we also add a BSD-specific dev-dependency on webpki-roots and update the real world verification suite to conditionally use the `Verifier::new_with_extra_roots` constructor to provide extra CA certs from webpki-roots.

It might be possible to fix the FreeBSD runner so that openssl-probe works (e.g. by `curl`ing a CA bundle into a different location, or setting the `SSL_CERT_FILE` env var) but this approach has the benefit of adding coverage for `new_with_extra_roots`.

Since GitHub actions doesn't offer FreeBSD runners we follow the Quinn project's lead and use `vmactions/freebsd-vm@v1` to run a FreeBSD VM on the runner, and our tests within the VM.

Resolves https://github.com/rustls/rustls-platform-verifier/issues/53

### DIFF
--- a/rustls-platform-verifier/Cargo.toml
+++ b/rustls-platform-verifier/Cargo.toml
@@ -35,7 +35,7 @@ base64 = { version = "0.21", optional = true } # Only used when the `cert-loggin
 jni = { version = "0.19", default-features = false, optional = true } # Only used during doc generation
 once_cell = { version = "1.9", optional = true } # Only used during doc generation.
 
-[target.'cfg(target_os = "linux")'.dependencies]
+[target.'cfg(all(unix, not(target_os = "android"), not(target_os = "macos"), not(target_os = "ios")))'.dependencies]
 rustls-native-certs = "0.6"
 once_cell = "1.9"
 webpki = { package = "rustls-webpki", version = "0.101", features = ["alloc", "std"] }
@@ -49,6 +49,10 @@ android_logger = { version = "0.13", optional = true } # Only used during testin
 
 [target.'cfg(target_arch = "wasm32")'.dependencies]
 once_cell = "1.9"
+webpki-roots = "0.25"
+
+# BSD targets require webpki-roots for the real-world verification tests.
+[target.'cfg(target_os = "freebsd")'.dev-dependencies]
 webpki-roots = "0.25"
 
 [target.'cfg(any(target_os = "macos", target_os = "ios"))'.dependencies]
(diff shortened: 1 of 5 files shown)


rustls-platform-verifier/src/tests/verification_mock/mod.rs
```diff
@@ -116,47 +111,47 @@ fn test_verification_without_mock_root() {
 // Verifies that our test trust anchor(s) are not trusted when `Verifier::new()`
 // is used.
 mock_root_test_cases! {
-    valid_no_stapling_dns [ any(windows, target_os = "android", target_os = "macos", target_os = "linux") ] => TestCase {
+    valid_no_stapling_dns [ any(windows, unix) ] => TestCase {
```
cpu, line 114: It's not clear from the diff but we still require the `cfg` portion of the `mock_root_test_cases!` macro for the `stapled_revoked_xxx` cases that are still gated to only Windows, Android or MacOS, they aren't _all_ `any(windows, unix)`.

rustls-platform-verifier/Cargo.toml
```diff
@@ -51,6 +51,10 @@ android_logger = { version = "0.13", optional = true } # Only used during testin
 once_cell = "1.9"
 webpki-roots = "0.25"
 
+# BSD targets require webpki-roots for the real-world verification tests.
+[target.'cfg(all(unix, not(target_os = "linux"), not(target_os = "android"), not(target_os = "macos"), not(target_os = "ios")))'.dev-dependencies]
```
complexspaces, line 55: I think that `cfg(target_os = "freebsd")` would be clearer here. If we add testing support for the other 2 BSD platforms Rust supports, we could add them specifically here.

rustls-platform-verifier/src/tests/verification_real_world/mod.rs
```diff
@@ -118,6 +118,35 @@ macro_rules! no_error {
 fn real_world_test<E: std::error::Error>(test_case: &TestCase<E>) {
     log::info!("verifying {:?}", test_case.expected_result);
 
+    // On BSD systems openssl-probe fails to find the system CA bundle,
+    // so we must provide extra roots from webpki-roots.
+    #[cfg(all(
```
complexspaces, line 123: Ditto on these two `cfg` blocks that were added as well.

complexspaces (changes requested):

rustls-platform-verifier/Cargo.toml
cpu: Good idea, fixed :-)

rustls-platform-verifier/src/tests/verification_real_world/mod.rs
cpu: Fixed here too.

complexspaces (approved):

PR #57, opened 2024-01-04
Fix incorrect platform-specific Verifier documentation

### LINKED ISSUE #56: Fix Verifier struct required feature flag rendering in docs.rs
@ralith pointed out that the `Verifier` struct is being rendered on `docs.rs` as "Available on Linux or WebAssembly only." - but that's not correct: each platform has its own version of this struct. At the time of writing this is visible [here](https://docs.rs/rustls-platform-verifier/latest/rustls_platform_verifier/struct.Verifier.html):

bad

Each of the platforms has its own `pub use` of the `Verifier` struct in the `verification` mod:
https://github.com/rustls/rustls-platform-verifier/blob/a586af91c780cec83625fbe05dcc4ce9f1d7c567/rustls-platform-verifier/src/verification/mod.rs#L4-L5

https://github.com/rustls/rustls-platform-verifier/blob/a586af91c780cec83625fbe05dcc4ce9f1d7c567/rustls-platform-verifier/src/verification/mod.rs#L10-L11

https://github.com/rustls/rustls-platform-verifier/blob/a586af91c780cec83625fbe05dcc4ce9f1d7c567/rustls-platform-verifier/src/verification/mod.rs#L16-L17

https://github.com/rustls/rustls-platform-verifier/blob/a586af91c780cec83625fbe05dcc4ce9f1d7c567/rustls-platform-verifier/src/verification/mod.rs#L22-L23

I'm not sure why the Linux/WASM one is being rendered in docs.rs but not the others. I did some experimenting with adjusting the `cfg` gates like we do in `lib.rs` so [Android specific features are displayed](https://github.com/rustls/rustls-platform-verifier/blob/a586af91c780cec83625fbe05dcc4ce9f1d7c567/rustls-platform-verifier/src/lib.rs#L11-L15) but wasn't able to find a solution.

You can reproduce this locally, but must do so with nightly rust and the `doc_auto_cfg` feature enabled with:
```
RUSTDOCFLAGS="--cfg docsrs" cargo +nightly doc
```

Putting this issue in the tracker so that it's acknowledged as a known issue waiting for a fix.

complexspaces:
This PR fixes the incorrect documentation on docs.rs pointed out in #56. The root cause of the problem was that when I originally wrote the Android documentation parts, I used the `doc_auto_cfg` feature instead of the `doc_cfg` by mistake. The former is too eager and assumes that anything with a `#[cfg()]` gate is only available under the gated conditions.

Instead of trying to get what we want with `doc_auto_cfg`, and especially since we only have one item that needs special documentation treatment, I've instead switched to use `doc_cfg` "manually". This approach was partially inspired by [how Tokio generates](https://github.com/tokio-rs/tokio/blob/master/tokio/src/macros/cfg.rs#L18-L26) their platform-specific documentation.

Documentation generated from a Windows system:
image

Documentation generated on a macOS system:
image

Finally, I fixed a broken doc warning in the README when referencing the `new_with_extra_roots` function. Since this function only exists on Linux right now, documentation built on Windows and macOS won't make that symbol available. I've chosen to link to docs.rs instead for that function at least until the other platforms gain that function too.

Closes #56

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,6 +41,7 @@ Users may wish to augment these certificates with [webpki-roots] using [`Verifie
 a system CA bundle is unavailable.
 
 [`ServerCertVerifierBuilder`]: https://docs.rs/rustls/latest/rustls/client/struct.ServerCertVerifierBuilder.html
+[`Verifier::new_with_extra_roots`]: https://docs.rs/rustls-platform-verifier/latest/rustls_platform_verifier/struct.Verifier.html#method.new_with_extra_roots
 [rustls-native-certs]: https://github.com/rustls/rustls-native-certs
 [openssl-probe]: https://github.com/alexcrichton/openssl-probe
 [webpki-roots]: https://github.com/rustls/webpki-roots
--- a/rustls-platform-verifier/src/lib.rs
+++ b/rustls-platform-verifier/src/lib.rs
@@ -1,4 +1,4 @@
-#![cfg_attr(docsrs, feature(doc_auto_cfg))]
+#![cfg_attr(docsrs, feature(doc_cfg))]
 #![doc = include_str!("../README.md")]
 #![warn(missing_docs)]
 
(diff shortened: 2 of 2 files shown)


cpu (approved):
Thank you for figuring this out! :trophy:

PR #65, opened 2024-01-24
Add a code sample showing how to use it

djc:
Was looking for this while writing on https://github.com/rustls/hyper-rustls/pull/253.

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,6 +52,25 @@ On most platforms, no setup should be required beyond adding the dependency via 
 rustls-platform-verifier = "0.1"
 ```
 
+To get a rustls `ClientConfig` configured to use the platform verifier use:
+
+```rust
+let config = rustls_platform_verifier::tls_config();
+```
+
+If you want to adapt the configuration, you can build the `ClientConfig` like this:
+
+```rust
+use std::sync::Arc;
+use rustls::ClientConfig;
+use rustls_platform_verifier::Verifier;
+
+let mut config = ClientConfig::builder()
+    .dangerous() // The `Verifier` we're using is actually safe
+    .with_custom_certificate_verifier(Arc::new(Verifier::new()))
+    .with_no_client_auth();
+```
+
 ### Android
 Some manual setup is required, outside of `cargo`, to use this crate on Android. In order to
 use Android's certificate verifier, the crate needs to call into the JVM. A small Kotlin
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -44,6 +44,25 @@ On most platforms, no setup should be required beyond adding the dependency via
...
+
+```rust
+use rustls::ClientConfig;
+use rustls_platform_verifier::Verifier;
+
+let mut config = ClientConfig::builder()
+    .with_safe_default_protocol_versions()
+    .dangerous() // The `Verifier` we're using is actually safe
+    .with_custom_certificate_verifier(Verifier::new())
+    .with_no_client_auth();
```
cpu, line 63: A couple small things:

1. The use of `with_safe_default_protocol_versions()` is a Rustls 0.21.x-ism that's implied by using `builder()` in 0.22. I think this is causing the CI task failures on this branch.

2. This block of code can be replaced with [`lib.rs`'s `tls_config()` fn](https://github.com/rustls/rustls-platform-verifier/blob/9188474a9c2ca04abc9acaac32b9c160661788f2/rustls-platform-verifier/src/lib.rs#L61-L66). That has the advantage of removing the need to call `dangerous` yourself with a clarifying comment. 

It might be worthwhile having an example of creating a customized `ClientConfig` (e.g for the case where you wanted to do client auth or something), but I think it should be placed after the more common example using `tls_config()`.

cpu (commented):
Thanks! I agree the README could provide more example code, however I have some feedback about the proposed addition.

README.md
djc: > It might be worthwhile having an example of creating a customized `ClientConfig` (e.g for the case where you wanted to do client auth or something), but I think it should be placed after the more common example using `tls_config()`.

I think that's what I did? 😄

cpu: > I think that's what I did? 😄 

Right, but I mean now the README _only_ has the uncommon getting started example where you want to customize the client config. 

I'm arguing that we should have that but only after showing an example of the more common "Just give me a client config that uses the platform verifier" example. That example is present today in `lib.rs`, but not in `README.md` where you're adding your customized example.

Sorry if that wasn't clear :laughing:

cpu: > Right, but I mean now the README only has the uncommon getting started example

:facepalm: Nope, I just can't read. I see it there. 

Sorry!

cpu (approved):

Issue #68, opened 2024-02-28
MacOS/Android AVD CI failures


cpu:
It looks like the MacOS CI jobs running the Android virtual device tests started failing on `main` ~yesterday ([one](https://github.com/rustls/rustls-platform-verifier/actions/runs/8069349802/job/22044760500), [two](https://github.com/rustls/rustls-platform-verifier/actions/runs/8085066172/job/22091793743)). Kicking the jobs doesn't resolve the failure, so it's probably not a transient flake.

```
> Task :rustls-platform-verifier:connectedDebugAndroidTest
Starting 3 tests on test(AVD) - 9

org.rustls.platformverifier.CertificateVerifierTests > runRealWorldTestSuite[test(AVD) - 9] FAILED 
	org.junit.ComparisonFailure: A test failed. Check the logs above for Rust panics. expected:<[success]> but was:<[test failed!]>
	at org.junit.Assert.assertEquals(Assert.java:115)
Tests on test(AVD) - 9 failed: There was 1 failure(s).
```

Unfortunately the mentioned Rust panic logs are not being passed through (similar to https://github.com/rustls/rustls-platform-verifier/pull/27).

complexspaces:
This seems like it would be a good time to try and move the Android jobs to Linux runners instead. We would need to [enable KVM](https://determinate.systems/posts/kvm-on-github-actions/) as well to keep decent performance.

cpu:
Sounds reasonable :+1:

cpu:
I was able to unbreak my local dev environment by pinning a specific android-ndk version (23.1.7779620) and when run locally the missing real world test suite panic logs do appear. It's the Let's Encrypt real world verification suite that seems to be failing in the Android VM (if we assume my local env matches what happens in CI). The relevant part was:

```
03-11 16:21:38.616  6704  6723 I rustls_platform_verif..: -----------------------------------------------------------------------------
03-11 16:21:38.616  6704  6731 I rustls_platform_verif..: verifying ref ID "letsencrypt.org" expected Ok(())
03-11 16:21:38.624  6704  6731 W rustls_platform_verif..: certificate was not trusted: java.security.cert.CertificateException: Chain validation failed
03-11 16:21:38.624  6704  6731 E rustls_platform_verif..: failed to verify TLS certificate: invalid peer certificate: UnknownIssuer
03-11 16:21:38.624  6704  6731 E rustls_platform_verif..: test panic: assertion `left == right` failed
[...]

PR #70, opened 2024-03-08
Update rustls to 0.23

Ralith:
Required for Quinn to update past rustls 0.21, so a release soon after would be appreciated. Also relaxes default feature requirements on rustls.

### DIFF
--- a/rustls-platform-verifier/Cargo.toml
+++ b/rustls-platform-verifier/Cargo.toml
@@ -26,29 +26,26 @@ dbg = []
 # by the platform's verifier.
 cert-logging = ["base64"]
 # Used for nicely documenting the Android-specific APIs. This feature is not stable.
-docsrs = ["jni", "once_cell"]
+docsrs = ["jni"]
 
 [dependencies]
 rustls = { version = "0.23", default-features = false, features = ["std"] }
 log = { version = "0.4" }
 base64 = { version = "0.21", optional = true } # Only used when the `cert-logging` feature is enabled.
 jni = { version = "0.19", default-features = false, optional = true } # Only used during doc generation
-once_cell = { version = "1.9", optional = true } # Only used during doc generation.
+once_cell = "1.9"
 
 [target.'cfg(all(unix, not(target_os = "android"), not(target_os = "macos"), not(target_os = "ios")))'.dependencies]
 rustls-native-certs = "0.7"
-once_cell = "1.9"
 webpki = { package = "rustls-webpki", version = "0.102", features = ["ring", "alloc", "std"] }
 
 [target.'cfg(target_os = "android")'.dependencies]
 rustls-platform-verifier-android = { path = "../android-release-support", version = "0.1.0" }
 jni = { version = "0.19", default-features = false }
 webpki = { package = "rustls-webpki", version = "0.102", features = ["ring", "alloc", "std"] }
-once_cell = "1.9"
 android_logger = { version = "0.13", optional = true } # Only used during testing.
 
 [target.'cfg(target_arch = "wasm32")'.dependencies]
-once_cell = "1.9"
 webpki-roots = "0.26"
 
 # BSD targets require webpki-roots for the real-world verification tests.
--- a/rustls-platform-verifier/Cargo.toml
+++ b/rustls-platform-verifier/Cargo.toml
@@ -29,7 +29,7 @@ cert-logging = ["base64"]
 docsrs = ["jni", "once_cell"]
 
 [dependencies]
-rustls = { version = "0.22.1", features = ["tls12", "logging"] }
+rustls = { version = "0.23", default-features = false, features = ["std", "ring"] }
 log = { version = "0.4" }
 base64 = { version = "0.21", optional = true } # Only used when the `cert-logging` feature is enabled.
 jni = { version = "0.19", default-features = false, optional = true } # Only used during doc generation
--- a/rustls-platform-verifier/Cargo.toml
+++ b/rustls-platform-verifier/Cargo.toml
@@ -19,7 +19,7 @@ crate-type = ["cdylib", "rlib"]
 # Enables a C interface to use for testing where `cargo` can't be used.
 # This feature is not stable, nor is the interface exported when it is enabled.
 # Do not rely on this or use it in production.
-ffi-testing = ["android_logger"]
+ffi-testing = ["android_logger", "rustls/ring"]
 # Enables APIs that expose lower-level verifier types for debugging purposes.
 dbg = []
 # Enables `log::debug` base64-encoded logging of all end-entity certificates processed
@@ -29,7 +29,7 @@ cert-logging = ["base64"]
 docsrs = ["jni", "once_cell"]
 
 [dependencies]
-rustls = { version = "0.23", default-features = false, features = ["std", "ring"] }
+rustls = { version = "0.23", default-features = false, features = ["std"] }
 log = { version = "0.4" }
 base64 = { version = "0.21", optional = true } # Only used when the `cert-logging` feature is enabled.
 jni = { version = "0.19", default-features = false, optional = true } # Only used during doc generation
@@ -63,6 +63,9 @@ security-framework-sys = { version = "2.4", features = ["OSX_10_14"] }
 
 [target.'cfg(windows)'.dependencies]
 winapi = { version = "0.3", features = ["wincrypt", "winerror"] }
+
+[dev-dependencies]
+rustls = { version = "0.23", default-features = false, features = ["ring"] }
 
 [package.metadata.docs.rs]
 rustdoc-args = ["--cfg", "docsrs"]
--- a/rustls-platform-verifier/src/tests/mod.rs
+++ b/rustls-platform-verifier/src/tests/mod.rs
@@ -61,3 +61,7 @@ pub(crate) fn verification_time() -> pki_types::UnixTime {
     // Monday, March 11, 2024 8:30:25 PM UTC
     pki_types::UnixTime::since_unix_epoch(Duration::from_secs(1_710_189_025))
 }
+
+fn ensure_global_state() {
+    let _ = rustls::crypto::ring::default_provider().install_default();
+}
--- a/rustls-platform-verifier/src/verification/apple.rs
+++ b/rustls-platform-verifier/src/verification/apple.rs
@@ -1,3 +1,5 @@
+use std::sync::Arc;
+
 use super::log_server_cert;
 use crate::verification::invalid_certificate;
 use core_foundation::date::CFDate;
@@ -43,7 +45,7 @@ pub struct Verifier {
     /// Testing only: The root CA certificate to trust.
     #[cfg(any(test, feature = "ffi-testing", feature = "dbg"))]
     test_only_root_ca_override: Option<Vec<u8>>,
-    default_provider: CryptoProvider,
+    default_provider: Arc<CryptoProvider>,
 }
 
 impl Verifier {
@@ -53,7 +55,9 @@ impl Verifier {
         Self {
             #[cfg(any(test, feature = "ffi-testing", feature = "dbg"))]
             test_only_root_ca_override: None,
-            default_provider: rustls::crypto::ring::default_provider(),
+            default_provider: rustls::crypto::CryptoProvider::get_default()
+                .expect("rustls default CryptoProvider not set")
+                .clone(),
         }
     }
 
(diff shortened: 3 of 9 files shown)


djc:
I think I tried this and it won't work on macOS...

Ralith:
This is difficult to test without CI working by default.

complexspaces:
@Ralith I added you with the `Triage` role on the repository (it grants minimal issue and PR permissions). I can remove it once you've merged this and GitHub doesn't see you as an "untrusted contributor" here anymore but hopefully it doesn't bug you for now.

Ralith:
Android CI failing:

```
> Task :rustls-platform-verifier:connectedDebugAndroidTest

[DDMLIB]: An unexpected packet was received before the handshake.
[DDMLIB]: An unexpected packet was received before the handshake.
[DDMLIB]: An unexpected packet was received before the handshake.
[DDMLIB]: An unexpected packet was received before the handshake.

> Task :rustls-platform-verifier:connectedDebugAndroidTest
Starting 3 tests on test(AVD) - 9

org.rustls.platformverifier.CertificateVerifierTests > runRealWorldTestSuite[test(AVD) - 9] FAILED 
	org.junit.ComparisonFailure: A test failed. Check the logs above for Rust panics. expected:<[success]> but was:<[test failed!]>
	at org.junit.Assert.assertEquals(Assert.java:115)
Tests on test(AVD) - 9 failed: There was 1 failure(s).
```
Not sure how to interpret that. There aren't any panics that I can see. "An unexpected packet was received" sounds like a rustls-side issue. There seems to be some sort of test result artifact that might have details, but it doesn't look human-readable.

zh-jq-b:
It seems to be the same issue as reported here https://github.com/rustls/rustls-platform-verifier/issues/68.

cpu:
> Not sure how to interpret that. There aren't any panics that I can see. "An unexpected packet was received" sounds like a rustls-side issue. 

I think these are unrelated noise. The same output has been happening [on passing builds in the past](https://github.com/rustls/rustls-platform-verifier/actions/runs/8053565882/job/21996258520). "ddmlib" looks like it's specific to the Android emulator<->Host debug bridge and so I think likely to be inconsequential to the Rustls tests.

> It seems to be the same issue as reported here https://github.com/rustls/rustls-platform-verifier/issues/68.

Yes, it is.

rustls-platform-verifier/src/verification/apple.rs
Ralith, line 1: Note changes here in particular. Not sure why this was hardcoding *ring* in the past.

Ralith:
If the only CI failure is #68, can we consider this unblocked?

rustls-platform-verifier/src/verification/apple.rs
cpu: Yup, exactly :+1:

cpu:
> If the only CI failure is https://github.com/rustls/rustls-platform-verifier/issues/68, can we consider this unblocked?

I don't know how @complexspaces feels but I think we wouldn't want to cut a release until `main` is passing CI so even if we merged this it would be blocked on #68. 

I've made some progress and think I have a fix. Will update #68 shortly

cpu:
> I've made some progress and think I have a fix. Will update https://github.com/rustls/rustls-platform-verifier/issues/68 shortly

I think cherry-picking in 98725337daa56ebe98db7877670b3886884539e8 from https://github.com/rustls/rustls-platform-verifier/pull/71 will sort the mac task out.

Ralith:
macos is hitting:

> rustls default CryptoProvider not set

which is from the `expect` I added on [`CryptoProvider::get_default()`](https://docs.rs/rustls/latest/rustls/crypto/struct.CryptoProvider.html#method.get_default) calls. This stopped happening on other platforms when the rustls `ring` feature was enabled. I'd expect to have to call `install_default` either everywhere or nowhere; having that requirement vary across platforms is baffling.

complexspaces:
That is extraordinarily strange. I tried comparing the output of `cargo tree --edges features` between macOS and Linux and nothing jumps out as "broken" in `rustls` direct dependency tree. The only difference is that there's no `rustls-webpki`.

Ralith:
Perhaps it's a test case ordering/global state issue. rustls tests suggest that `None` is the expected return value even when the feature is enabled:

https://github.com/rustls/rustls/blob/0398ac50fe18521da038aa39b55eec6c83450523/rustls/tests/process_provider.rs#L47-L58

I'll try setting defaults explicitly.

complexspaces:
It looks like its back to the Android failures 🎉 If you rebase on `main` Ci should be green at least. I merged #71 a few minutes ago.

rustls-platform-verifier/src/verification/windows.rs
```diff
@@ -438,7 +441,9 @@ impl Verifier {
     pub(crate) fn new_with_fake_root(root: &[u8]) -> Self {
         Self {
             test_only_root_ca_override: Some(root.into()),
-            default_provider: rustls::crypto::ring::default_provider(),
+            default_provider: rustls::crypto::CryptoProvider::get_default()
```
djc, line 444: All these constructors should probably document that they're relying on the process-global crypto provider set via rustls?

rustls-platform-verifier/src/tests/mod.rs
```diff
@@ -61,3 +61,10 @@ pub(crate) fn verification_time() -> pki_types::UnixTime {
     // Monday, March 11, 2024 8:30:25 PM UTC
     pki_types::UnixTime::since_unix_epoch(Duration::from_secs(1_710_189_025))
 }
+
+fn ensure_global_state() {
```
djc, line 65: I think the `cfg(test)` guard can be outside the function?

cpu (commented):
Reviewing the diff I don't have any feedback above what djc has already flagged. Thanks again for picking this up Ralith :tada:

rustls-platform-verifier/src/tests/mod.rs
complexspaces: ++ on this. We shouldn't need the guard at all because the `tests` module should be under the same thing already.

Ralith: That's what I thought, too. However, the `tests` module is gated by `#[cfg(any(test, feature = "ffi-testing"))]`, and whatever builds it with `ffi-testing` in CI isn't a test build, so it doesn't get the *ring* dev-dependency, so the build fails. I think this is also related to the remaining failure, in which the android tests presumably don't set up a default provider. Haven't actually found the android tests or worked out a strategy to address that yet. Maybe we need some sort of automatic fallback like `rustls` seems to have.

complexspaces: That looks like it it actually, good catch. With the improved Android logging I saw it was throwing the missing crypto provider error for every test. Locally I updated the `ffi-testing` feature and made sure to call `ensure_global_state` in the Android tests. After that everything worked. Here's the changes if you'd like to commit them:
```diff
diff --git a/rustls-platform-verifier/Cargo.toml b/rustls-platform-verifier/Cargo.toml
index cdffcdf..dc2f369 100644
--- a/rustls-platform-verifier/Cargo.toml
+++ b/rustls-platform-verifier/Cargo.toml
@@ -19,7 +19,7 @@ crate-type = ["cdylib", "rlib"]
 # Enables a C interface to use for testing where `cargo` can't be used.
 # This feature is not stable, nor is the interface exported when it is enabled.
 # Do not rely on this or use it in production.
-ffi-testing = ["android_logger"]
+ffi-testing = ["android_logger", "rustls/ring"]
 # Enables APIs that expose lower-level verifier types for debugging purposes.
 dbg = []
 # Enables `log::debug` base64-encoded logging of all end-entity certificates processed
diff --git a/rustls-platform-verifier/src/tests/ffi.rs b/rustls-platform-verifier/src/tests/ffi.rs
index bf18a01..6a4a5af 100644
--- a/rustls-platform-verifier/src/tests/ffi.rs
+++ b/rustls-platform-verifier/src/tests/ffi.rs
@@ -47,6 +47,7 @@ mod android {
                     .with_filter(log_filter),
             );
             crate::android::init_hosted(env, cx).unwrap();
+            crate::tests::ensure_global_state();
             std::panic::set_hook(Box::new(|info| {
                 let msg = if let Some(msg) = info.payload().downcast_ref::<&'static str>() {
                     msg
diff --git a/rustls-platform-verifier/src/tests/mod.rs b/rustls-platform-verifier/src/tests/mod.rs
index 63e6490..16ff02a 100644
--- a/rustls-platform-verifier/src/tests/mod.rs
+++ b/rustls-platform-verifier/src/tests/mod.rs
@@ -62,9 +62,6 @@ pub(crate) fn verification_time() -> pki_types::UnixTime {
     pki_types::UnixTime::since_unix_epoch(Duration::from_secs(1_710_189_025))
 }
 
-fn ensure_global_state() {
-    #[cfg(test)]
-    {
-        _ = rustls::crypto::ring::default_provider().install_default();
-    }
+pub(crate) fn ensure_global_state() {
+    let _ = rustls::crypto::ring::default_provider().install_default();
 }
```

rustls-platform-verifier/src/verification/windows.rs
Ralith: I think proceeding with this design would be user-hostile, as a typical use is likely to panic at runtime for users who haven't explicitly installed a default, which isn't normally required by rustls. I think a better solution would be for rustls to expose `CryptoProvider::get_default_or_install_from_crate_features`, which would let us Just Work as well as rustls itself does (although I have mixed feelings about how that breaks if multiple providers are enabled). 

We could also reimplement the implicit default logic with local features in this crate, but that would put users at higher risk of accidentally enabling multiple providers (and potentially breaking any part of their code that relies on implicit defaults) if they don't take care to match features on rustls and this crate.

djc: I think this crate should probably just take an Arc<CryptoProvider> as an argument.

Ralith: That works, though it undermines the convenience of rustls's implicit defaults if an idiomatic user has to go and deal with explicit provider handling for the verifier anyway.

complexspaces: > My other thought on this is: perhaps we should aim for windows.rs to use CryptVerifyMessageSignature (or similar) and for apple.rs to use SecKeyVerifySignature?

I have thought about this before, but I'm still not sure I like the idea. It adds another "split" in the library's functionality set, especially on Windows where registry controls can influence what cryptographic operations the OS prefers/will use. This would make debugging harder, for example. So far I have only seen one cryptographic issue come in via bug report at work. Everything else has been missing trust roots, malformed X.509, etc.

Ralith: > It adds another "split" in the library's functionality set

Isn't exposing platform-specific behavior the whole point of this lib?

> I think it might be cleaner if this operates at the level of a "parent" Arc<dyn ServerCertVerifier> rather than a CryptoProvider.

This makes a lot of sense to me.

> But unfortunately right now WebPkiServerVerifier::builder() does require a non-empty set of roots, which makes it annoying to reuse for this purpose.

Could we relax this, or introduce an alternative verifier in rustls that has an empty root set but implements the other methods? Practically speaking this amounts to creatively delegating to rustls's `get_default_or_install_from_crate_features`, which I think is the best user experience.

Ralith:
I've added a commit that makes calls to `CryptoProvider::get_default()` lazy, which should ensure a good user experience without hard-coding a specific provider or breaking the API.

Long-term, I think it would be better for rustls to expose `CryptoProvider::get_default_or_install_from_crate_features` directly, since its use in e.g. `WebPkiServerVerifier` seems to be a tacit admission that it's the most user-friendly way for library code to get a provider. However, this change will let rustls-platform-verifier update and unblock downstream users like Quinn immediately without regressing usability.

rustls-platform-verifier/Cargo.toml
```diff
@@ -29,7 +29,7 @@ cert-logging = ["base64"]
 docsrs = ["jni", "once_cell"]
 
 [dependencies]
-rustls = { version = "0.22.1", features = ["tls12", "logging"] }
```
cpu, line 32: Was it intentional to drop the `tls12` feature activation?

cpu, line 32: Was it intentional to drop the `tls12` feature activation?

cpu (approved):

rustls-platform-verifier/Cargo.toml
Ralith: Yep, very intentional. For example, Quinn users will typically not want TLS 1.2 support built. It didn't actually seem to be needed, anyway.

cpu: Makes sense, just wanted to check explicitly 👍 Thanks!

Lewiscowles1986:
Please merge

ibigbug:
are we good to go with this? this seems to be the only blocker of https://github.com/quinn-rs/quinn/pull/1715 ?

complexspaces (approved):
Sorry for the delay, LGTM.

PR #71, opened 2024-03-11
tests: refresh LE real world verification test case, more logging

cpu:
As [described in #68](https://github.com/rustls/rustls-platform-verifier/issues/68#issuecomment-1989444096) updating the end-entity certificate used in the Let's Encrypt real world verification integration test suite seems to fix the CI failure in the Android tasks that creeped into `main` ~2 weeks ago.

There are remaining mysteries to address, but this should be enough to fix CI while we dig deeper.


cpu:
> 14 successful checks

:green_heart: :heavy_check_mark:

complexspaces:
This seems good to me, but I want to quickly try getting #72 working so we have these logs again first.

complexspaces (approved):

PR #72, opened 2024-03-12
Update Android test action and fix outstanding CI issues

complexspaces:
This PR performs some maintenance on our Android testing infrastructure. Notably it:
1. Updates our Android runner action to try and pick up on bugfixes.
2. Switches the job to run on Linux instead for speed. We enable KVM on the runner as well since the Android emulator needs virtualization and that needs to be fast.
3. Moves the Android job into its own YAML area so it doesn't get confused with the host OS's tests anymore.
4. Fixes the log output and artifact upload locations

Relates to #68

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -116,16 +116,30 @@ jobs:
       - name: Test (${{ matrix.os }})
         run: cargo test
 
-      - name: Setup Android test environment
+  test_android:
+    name: "Test (Android)"
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      # Turn on Linux KVM features/support for faster Android emulation.
+      # References:
+      # - https://github.com/DeterminateSystems/nix-installer-action/blob/de22e16c4711fca50c816cc9081563429d1cf563/src/main.ts#L756
+      # - https://github.com/ReactiveCircus/android-emulator-runner#running-hardware-accelerated-emulators-on-linux-runners
+      - name: Enable KVM
+        run: |
+          echo 'KERNEL=="kvm", GROUP="kvm", MODE="0666", OPTIONS+="static_node=kvm"' | sudo tee /etc/udev/rules.d/99-kvm4all.rules
+          sudo udevadm control --reload-rules
+          sudo udevadm trigger --name-match=kvm
+
+      - name: Setup Java
         uses: actions/setup-java@v2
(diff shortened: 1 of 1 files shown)


complexspaces:
This PR is ready to go. We are now properly writing device/test logs to stdout and uploading the correct log output path:

image

.github/workflows/ci.yml
```diff
@@ -116,16 +116,26 @@ jobs:
...
+    name: "Test (Android)"
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Enable KVM
+        run: |
+          echo 'KERNEL=="kvm", GROUP="kvm", MODE="0666", OPTIONS+="static_node=kvm"' | sudo tee /etc/udev/rules.d/99-kvm4all.rules
+          sudo udevadm control --reload-rules
+          sudo udevadm trigger --name-match=kvm
```
cpu, line 129: This is sufficiently magical seeming it might be worth leaving a pointer comment to [the android-emulator-runner README](https://github.com/ReactiveCircus/android-emulator-runner?tab=readme-ov-file#running-hardware-accelerated-emulators-on-linux-runners) where it documents this setup. WDYT?

.github/workflows/ci.yml
```diff
@@ -146,22 +156,23 @@ jobs:
             env | grep '^JAVA'
             touch emulator.log
             chmod 770 emulator.log
-            adb logcat >> emulator.log &
+            adb logcat -c
```
cpu, line 159: minor nit: I'd prefer long options here so it takes less time for folks unfamiliar with adb to grok what's going on:

```suggestion
            adb logcat --clear
```

On that note I was surprised we wanted to clear the log here, but the `--help` output says:

> Clear (flush) the entire log and exit.

I suppose it's the "flushing" behaviour we're interested in?

cpu (approved):
Thank you for figuring this out! It'll be a really nice quality of life improvement.

.github/workflows/ci.yml
complexspaces: I agree, good callout. I've updated this with two references to elsewhere about how to implement this.

.github/workflows/ci.yml
complexspaces: That's correct. I figured it was better to flush these ahead-of-time just in case we implement AVD caching in the future to speed CI up.

cpu (approved):

PR #73, opened 2024-03-12
proj: add initial Nix flake support

cpu:
The tooling required to work in this repo can be a little bit challenging to set up (largely as a side effect of supporting multiple platforms/languages, each with their own tooling requirements).

In an effort to make this easier and more reproducible this commit adds the start of a [Nix flake](https://zero-to-nix.com/concepts/flakes).

Some starting points (assuming you've set up `nix`, which is outside of the scope of this PR):

* `nix build .` - builds the platform verifier Rust library in a release configuration using stable Rust.
* `nix build .#rustls-platform-verifier-dbg` - same as above, but with the "dbg", "base64", and "docsrs" Cargo features enabled.
* `nix develop` - enter a development shell offering a nightly rust toolchain, Android studio, Android SDK/NDK, Java, and Maven tools for building the Android project and using the release tooling.
* `nix develop .#stable` - same as above, but with the stable Rust toolchain.
* `nix develop .#msrv` - same as above, but with the project MSRV Rust toolchain.

There's also an `android-emu` output for creating an Android device emulator that can be used to run unit tests:

```
# Terminal 1 - Android emulator
nix develop
nix build .#android-emu
./result/bin/run-test-emulator&
adb devices
adb logcat
```

```
# Terminal 2 - Unit tests
nix develop
pushd android && ./gradlew connectedAndroidTest
```

Future work:

* The NDK version is presently pinned to a 23.x version to avoid a clang libc++ error. This should be resolved so we can use 24.x (Some [debugging notes here](https://gist.github.com/cpu/d67e41030196d3fe7f5c00fe50aa2f28))
* Support for additional platforms. I believe most of this will work on other systems/targets but I've only tested on x86_64 so have left out additional support for now
* ~It should be possible to configure an Android emulator, and to run the unit tests within the emulator, without using Android studio or manually configuring an emulator.~ I added an output to create an emulator device.
* ~I'm not sure whether there's interest in CI integration. It might be a good idea to add a `nix build` step somewhere to try and prevent bitrot.~ I added a CI task.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -1,0 +1,130 @@
+{
+  inputs = {
+    nixpkgs.url = "github:NixOS/nixpkgs/nixos-23.11";
+    flake-parts.url = "github:hercules-ci/flake-parts";
+    rust-overlay.url = "github:oxalica/rust-overlay";
+  };
+
+  outputs = inputs:
+    inputs.flake-parts.lib.mkFlake { inherit inputs; } {
+      # TODO(XXX): in theory this flake could support aarch64-linux,
+      #   x86_64-darwin and aarch64-darwin, but it is untested.
+      systems = [ "x86_64-linux" ];
+      perSystem = { config, self', pkgs, lib, system, ... }:
+        let
+          buildToolsVersion = "30.0.3";
+          platformVersion = "33";
+          rustTargets = [ "${system}-android" ];
+          abi = (lib.systems.elaborate system).linuxArch;
+
+          devDeps = with pkgs; [
+            android-sdk
+            android-studio
+            maven
+            jdk11 # Matched to CI setup-java task's java-version
+            der-ascii
+          ];
+
+          android-comp = pkgs.androidenv.composeAndroidPackages {
+            buildToolsVersions = [ buildToolsVersion ];
+            platformVersions = [ platformVersion ];
+            abiVersions = [ abi ];
+
+            # Note: Pinned presently to NDK 23 specifically to workaround an issue
+            # with the bundled clang missing libc++ in NDK 24. We can't use NDK
+            # 22, as it's too old for cargo-ndk.
+            ndkVersion = "23.1.7779620";
+
+            systemImageTypes = [ "default" ];
+
+            includeNDK = true;
+            includeEmulator = true;
+            includeSystemImages = true;
+          };
+
+          # Note: additional flags can be provided to emulator through
+          #   the $NIX_ANDROID_EMULATOR_FLAGS env var.
+          android-emu = pkgs.androidenv.emulateApp {
+            name = "emulate-PlatformVerifier";
+            platformVersion = platformVersion;
+            abiVersion = abi;
+            systemImageType = "default";
+
+            # Note: Depending on your hardware you may wish to enable or disable
+            # this option.
+            enableGPU = false;
+          };
+
+          android-sdk = android-comp.androidsdk;
+          android-sdk-root = "${android-sdk}/libexec/android-sdk";
+
+          verifierCargoToml = builtins.fromTOML
+            (builtins.readFile ./rustls-platform-verifier/Cargo.toml);
+          msrv = verifierCargoToml.package.rust-version;
+
+          verifierPackage = features:
+            (pkgs.makeRustPlatform {
+              cargo = pkgs.rust-bin.stable.latest.minimal;
+              rustc = pkgs.rust-bin.stable.latest.minimal;
+            }).buildRustPackage {
+              inherit (verifierCargoToml.package) name version;
+              src = ./.;
+              buildAndTestSubdir = "rustls-platform-verifier";
+              cargoLock.lockFile = ./Cargo.lock;
+              buildFeatures = features;
+              doCheck = false; # Some tests require networking
+            };
+
+          mkDevShell = rustc:
+            pkgs.mkShell {
+              ANDROID_HOME = "${android-sdk-root}";
+              ANDROID_SDK_ROOT = "${android-sdk-root}";
+              ANDROID_NDK_ROOT = "${android-sdk-root}/ndk-bundle";
+              JAVA_HOME = "${pkgs.jdk11}";
+              # Note: It's important to set this so that gradle uses the correct
+              #   aapt2 binary.
+              GRADLE_OPTS =
+                "-Dorg.gradle.project.android.aapt2FromMavenOverride=${android-sdk-root}/build-tools/${buildToolsVersion}/aapt2";
+              shellHook = ''
+                export RUST_SRC_PATH=${pkgs.rustPlatform.rustLibSrc}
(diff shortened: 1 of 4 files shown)


flake.nix
```diff
@@ -0,0 +1,115 @@
...
+          rustTargets = [ "x86_64-linux-android" ];
+
+          devDeps = with pkgs; [
+            android-sdk
+            android-studio
+            maven
+            jdk11 # Matched to CI setup-java task's java-version
+          ];
+
+          android-comp = pkgs.androidenv.composeAndroidPackages {
```
cpu, line 26: Relevant docs: https://nixos.org/manual/nixpkgs/stable/#android

flake.nix
```diff
@@ -0,0 +1,115 @@
...
+
+  outputs = inputs:
+    inputs.flake-parts.lib.mkFlake { inherit inputs; } {
+      # TODO(XXX): in theory this flake could support aarch64-linux,
+      #   x86_64-darwin and aarch64-darwin, but it is untested.
+      systems = [ "x86_64-linux" ];
+      perSystem = { config, self', pkgs, lib, system, ... }:
+        let
+          buildToolsVersion = "30.0.3";
+          platformVersion = "33";
```
cpu, line 16: I can't remember how I landed on these versions - probably cribbing from example docs?

In CI we use a slightly older platform version (28 - "Android 9/Pie") and I'm not sure what build tool versions are in use.

djc:
I think @Ralith has some Nix experience, maybe he can help review?

flake.nix
```diff
@@ -0,0 +1,115 @@
...
+          ];
+
+          android-comp = pkgs.androidenv.composeAndroidPackages {
+            buildToolsVersions = [ buildToolsVersion ];
+            platformVersions = [ platformVersion ];
+            abiVersions = [ "x86_64" ];
+
+            # Matched to CI's Run Android Tests task - we want google_apis to
+            # ensure network access for real-world test suite.
+            systemImageTypes = [ "google_apis" ];
```
cpu, line 33: Leaving myself a note to rm this after https://github.com/rustls/rustls-platform-verifier/pull/74 lands

flake.nix
```diff
@@ -0,0 +1,115 @@
...
+            config = {
+              # Allow unfree packages and agree to the Android SDK terms of service.
+              # Review https://developer.android.com/studio/terms before use.
+              allowUnfree = true;
+              android_sdk.accept_license = true;
+            };
+          };
+
+          # Base library.
+          packages.rustls-platform-verifier = (verifierPackage "");
```
Ralith, line 95:
```suggestion
          packages.rustls-platform-verifier = (verifierPackage []);
```

Ralith (approved):
I don't have a lot of experience with flakes specifically, but this looks generally reasonable. I think we should have CI coverage for it.

flake.nix
cpu: Good catch, thanks. Fixed.

cpu:
> I don't have a lot of experience with flakes specifically, but this looks generally reasonable.

Thanks for taking a look :-)

>  I think we should have CI coverage for it.

I pushed a simple CI job :+1:

Ralith (approved):

cpu:
> It should be possible to configure an Android emulator, and to run the unit tests within the emulator, without using Android studio or manually configuring an emulator. 

I pushed one last revision to add an emulator image output. The PR description is updated with some quick usage instructions. This is sufficient for being able to build the project and run the Android tests in an emulator without doing any manual setup in Android studio.

Going to call this done for now and pick up on the NDK version pinning the next time I have a chance to fiddle with this.

flake.nix
cpu: Fixed and pushed, but it seems like there's some lag so the commit hasn't shown up in the UI yet. I'm logging out for the evening but will follow-up tomorrow.

flake.nix
```diff
@@ -0,0 +1,132 @@
...
+      # TODO(XXX): in theory this flake could support aarch64-linux,
+      #   x86_64-darwin and aarch64-darwin, but it is untested.
+      systems = [ "x86_64-linux" ];
+      perSystem = { config, self', pkgs, lib, system, ... }:
+        let
+          buildToolsVersion = "30.0.3";
+          platformVersion = "33";
+          rustTargets = [ "x86_64-linux-android" ];
+
+          devDeps = with pkgs; [
```
complexspaces, line 19: Maybe add [`der-ascii`](https://search.nixos.org/packages?channel=23.11&show=der-ascii&from=0&size=50&sort=relevance&type=packages&query=der-ascii) here too since its packaged and up-to-date?

cpu: Sure, good idea :+1: 

Added.

flake.nix
```diff
@@ -0,0 +1,129 @@
...
+  outputs = inputs:
+    inputs.flake-parts.lib.mkFlake { inherit inputs; } {
+      # TODO(XXX): in theory this flake could support aarch64-linux,
+      #   x86_64-darwin and aarch64-darwin, but it is untested.
+      systems = [ "x86_64-linux" ];
+      perSystem = { config, self', pkgs, lib, system, ... }:
+        let
+          buildToolsVersion = "30.0.3";
+          platformVersion = "33";
+          rustTargets = [ "x86_64-linux-android" ];
```
cpu, line 17: With fresh eyes it seems silly to not derive this from the system` arg given to the `perSystem` func by `mkFlake`. Will try to fix.

flake.nix
```diff
@@ -0,0 +1,129 @@
...
+            android-studio
+            maven
+            jdk11 # Matched to CI setup-java task's java-version
+            der-ascii
+          ];
+
+          android-comp = pkgs.androidenv.composeAndroidPackages {
+            buildToolsVersions = [ buildToolsVersion ];
+            platformVersions = [ platformVersion ];
+            abiVersions = [ "x86_64" ];
```
cpu, line 30: Ditto for these `abiVersions` args: should be derived from `system`.

flake.nix
cpu: Fixed.

flake.nix
cpu: Fixed.

cpu:
> CI / Clippy (-Zbuild-std) (pull_request) Failing after 14s
> cargo deny / audit (pull_request) Failing after 23s

Unrelated to this branch, resolving these in https://github.com/rustls/rustls-platform-verifier/pull/91

@complexspaces Are you thinking that you'd like to do some further reviewing/testing of this branch or could it be merged soon?

complexspaces (approved):
I've gotten slightly better (only slightly 😄) with Nix since I last looked at this, but AFAICT this looks fine. I'm looking forward to trying this out when time permits.

cpu:
>  cpu force-pushed the cpu-start-nix-flake_dev branch from 25440f2 to 58a085c 

Rebased on main to pick up the CI fixes. Will merge once this is all green.

Issue #59, opened 2024-01-05
Android verifier doesn't consider supplied date for revocation or intermediate roots


complexspaces:
Today in the Android verifier, [we only check the end-entity](https://github.com/rustls/rustls-platform-verifier/blob/main/android/rustls-platform-verifier/src/main/java/org/rustls/platformverifier/CertificateVerifier.kt#L206-L213)'s validity against the timestamp provided by `rustls`. The revocation status and intermediate CA's validity uses the system's current timestamp. While this doesn't result in any security issues, it makes consistent testing harder.

To fix this, we need to use a use the [`setDate`](https://developer.android.com/reference/java/security/cert/PKIXParameters#setDate(java.util.Date)) method of `PKIXParameters`. We already construct one of these for checking revocation, but we're using `TrustManager.checkServerTrusted` for the rest of the chain. Android's trust manager [defaults to using the current date](https://github.com/google/conscrypt/blob/932d847bffbc65fbaf649a272607bea0cb01fb90/common/src/main/java/org/conscrypt/TrustManagerImpl.java#L594), with no way to override it here.

Modifying the logic to do everything with one verifier will be tricky, if not infeasible. This is because the Conscrypt TrustManager checks for [Android-specific CA blocklisting](https://github.com/google/conscrypt/blob/932d847bffbc65fbaf649a272607bea0cb01fb90/common/src/main/java/org/conscrypt/TrustManagerImpl.java#L534) and CT/pinning requirements, and we definitely want to keep that functionality so Android matches what Apple and Windows are doing. As is, I'm not sure how to solve the intermediate time period problem. Even if we tried doing two validations with each API set, we most likely wouldn't (unless the system already knew about every intermediate and root) see any blocklist errors because Conscrypt checks validity _early_. From reading through the implementation, my conclusion is that [while including intermediates provided by the server](https://github.com/google/conscrypt/blob/932d847bffbc65fbaf649a272607bea0cb01fb90/common/src/main/java/org/conscrypt/TrustManagerImpl.java#L582-L583) an expired one would fail early before it had a chance to check it against the blocklist :( To fully resolve this, and get the best of what the platform has to offer we really need the ability to pass a timestamp into `checkServerTrusted`.

With all that said, a partial and feasible fix is to just start setting the datetime for our revocation checks and use the fact intermediates don't expire often to hopefully treat this a non-issue in practice.

cpu:
[...]

PR #75, opened 2024-03-12
android: more detail for test config verify exception

cpu:
Recently CI started failing the real world verification suite on Android (https://github.com/rustls/rustls-platform-verifier/issues/68). The root cause was the vendored Let's Encrypt end entity certificate expiring. Figuring this out was hard for two main reasons:

1. CI was eating the Android logs we needed to see the error message. This has since been fixed (https://github.com/rustls/rustls-platform-verifier/pull/72).

2. The Android verifier was surfacing the expiration as an unknown issuer error with a generic "Chain validation failed" message.

This branch attempts to fix item 2 by surfacing a more relevant error message.

Unfortunately doing so is a little bit annoying. The `checkServerTrusted` call throws an exception with ~3 layers of wrapping before you get at the `CertificateExpiredException` that's the root cause.

Since non-test configurations should never have an expired exception thrown at this stage (recall we check this explicitly earlier in processing), we gate the more complicated exception "digging" based on `BuildConfig.TEST`.

This should be a good balance of:

* not changing existing behaviour, or introducing complexity for normal validation
* surfacing a better error message when test certificates expire and start to break CI.

Before this fix running the verification test suite back at 6cd02326f1be869b4a6732f262bb04c4958ecf3b before 15d487a9c6e4bf23bc54ac2450d074c10f63eb69 landed would produce output like:

```
03-12 15:05:38.900 18229 18250 W rustls_platform_verif..: certificate was not trusted: java.security.cert.CertificateException: Chain validation failed
03-12 15:05:38.900 18229 18250 E rustls_platform_verif..: failed to verify TLS certificate: invalid peer certificate: UnknownIssuer
03-12 15:05:38.900 18229 18250 E rustls_platform_verif..: test panic: assertion `left == right` failed
03-12 15:05:38.900 18229 18250 E rustls_platform_verif..:   left: Err(InvalidCertificate(UnknownIssuer))
03-12 15:05:38.900 18229 18250 E rustls_platform_verif..:  right: Ok(())
03-12 15:05:38.901 18229 18248 E rustls_platform_verif..: real world: test failed
```

Afterwards it logs output like:

```
03-12 15:06:05.297 18482 18503 E rustls_platform_verif..: failed to verify TLS certificate: invalid peer certificate: Expired
03-12 15:06:05.297 18482 18503 E rustls_platform_verif..: test panic: assertion `left == right` failed
03-12 15:06:05.297 18482 18503 E rustls_platform_verif..:   left: Err(InvalidCertificate(Expired))
03-12 15:06:05.297 18482 18503 E rustls_platform_verif..:  right: Ok(())
03-12 15:06:05.298 18482 18501 E rustls_platform_verif..: real world: test failed
```

### DIFF
--- a/android/rustls-platform-verifier/src/main/java/org/rustls/platformverifier/CertificateVerifier.kt
+++ b/android/rustls-platform-verifier/src/main/java/org/rustls/platformverifier/CertificateVerifier.kt
@@ -253,6 +253,30 @@ internal object CertificateVerifier {
         val validChain = try {
             trustManager.checkServerTrusted(certificateChain.toTypedArray(), authMethod, serverName)
         } catch (e: CertificateException) {
+            // In test configurations we may see `checkServerTrusted` fail once vendored test
+            // certificates pass their expiry date. We try to avoid that by using a fixed
+            // verification time when calling `endEntity.checkValidity` above, however we can't
+            // fix the time for the `checkServerTrusted` call.
+            //
+            // To make diagnosing CI test failures easier we try to find the root cause of
+            // checkServerTrusted failing, returning a different `StatusCode` as appropriate.
+            if (BuildConfig.TEST) {
+                var rootCause: Throwable? = e
+                while (rootCause?.cause != null && rootCause.cause != rootCause) {
+                    rootCause = rootCause.cause
+                }
+                return when (rootCause) {
+                    is CertificateExpiredException, is CertificateNotYetValidException -> VerificationResult(
+                        StatusCode.Expired,
+                        rootCause.toString()
+                    )
+
+                    else -> VerificationResult(StatusCode.UnknownCert, rootCause.toString())
+                }
+            }
+            // In non-test configurations we should have caught expiry errors earlier and
+            // can simply return an unknown cert error without digging through the exception
+            // cause chain.
             return VerificationResult(StatusCode.UnknownCert, e.toString())
         }
 


android/rustls-platform-verifier/src/main/java/org/rustls/platformverifier/CertificateVerifier.kt
```diff
@@ -253,6 +253,30 @@ internal object CertificateVerifier {
...
+            // verification time when calling `endEntity.checkValidity` above, however we can't
+            // fix the time for the `checkServerTrusted` call.
+            //
+            // To make diagnosing CI test failures easier we try to find the root cause of
+            // checkServerTrusted failing, returning a different `StatusCode` as appropriate.
+            if (BuildConfig.TEST) {
+                var rootCause: Throwable? = e
+                while (rootCause?.cause != null && rootCause.cause != rootCause) {
+                    rootCause = rootCause.cause
+                }
```
cpu, line 267: I cribbed this from [this baeldung post](https://www.baeldung.com/java-exception-root-cause), and then tried my best as a non-Kotlin programmer to make it Kotlin-ish :laughing:

complexspaces: I believe this can be slightly cleaned up to make it more clear `rootCause` will never actually be `null`:
```diff
diff --git a/android/rustls-platform-verifier/src/main/java/org/rustls/platformverifier/CertificateVerifier.kt b/android/rustls-platform-verifier/src/main/java/org/rustls/platformverifier/CertificateVerifier.kt
index 79421ed..0290956 100644
--- a/android/rustls-platform-verifier/src/main/java/org/rustls/platformverifier/CertificateVerifier.kt
+++ b/android/rustls-platform-verifier/src/main/java/org/rustls/platformverifier/CertificateVerifier.kt
@@ -261,9 +261,11 @@ internal object CertificateVerifier {
             // To make diagnosing CI test failures easier we try to find the root cause of
             // checkServerTrusted failing, returning a different `StatusCode` as appropriate.
             if (BuildConfig.TEST) {
-                var rootCause: Throwable? = e
-                while (rootCause?.cause != null && rootCause.cause != rootCause) {
-                    rootCause = rootCause.cause
+                var rootCause: Throwable = e
+                while (rootCause.cause != rootCause) {
+                    rootCause.cause?.let {
+                        rootCause = it
+                    }
                 }
                 return when (rootCause) {
                     is CertificateExpiredException, is CertificateNotYetValidException -> VerificationResult(
```

cpu:
I'm starting to wonder if instead of these gymnastics we should just back out the `time` argument being fed through to the `CertificateVerifier` until we work out a broader solution to https://github.com/rustls/rustls-platform-verifier/issues/59. That would let the early `checkValidity` call surface the better error early on.

android/rustls-platform-verifier/src/main/java/org/rustls/platformverifier/CertificateVerifier.kt
cpu: Good call, fixed. Thanks!

cpu: I had to revert this change because it causes an infinite loop that binds up the test suite :skull: 

CI has been running the Android test suite with this commit applied for 1h+

cpu:
> I'm starting to wonder if instead of these gymnastics we should just back out the time argument being fed through to the CertificateVerifier until we work out a broader solution to https://github.com/rustls/rustls-platform-verifier/issues/59.

@complexspaces WDYT :point_up: vs this branch? In a perfect world we'd fix #59 instead of doing either but I don't think I'll personally be able to find time to work on that solution and it would be nice to have CI report the correct failure before it happens on April 26th (and every 90d afterwards).

cpu:
>  What do you think about just merging this to keep CI nicer and maintain the "good" errors most end users will see?

That works for me :-) If you have a chance to put a review on this branch I'll merge when ready.

complexspaces (approved):

cpu:
Thanks!

PR #76, opened 2024-03-22
chore: update rusttls to 0.23.2

Lewiscowles1986:
Hello, thanks for authoring this code.

I want to upgrade this to 0.23.2 in https://github.com/awslabs/llrt/pull/280 so that there are not two versions of a crate installed.

That requires a new https://github.com/rustls/pki-types
That requires a new https://github.com/rustls/hyper-rustls
That requires your package to update it's rustls

So thank you for writing this, but I would like to increase this dependency version and if possible ask for a new release 🙏


Lewiscowles1986:
I Have no idea why so many files have been updated in the lockfile. The only commands I've issued are
`cargo update rustls --precise 0.23.2`

flisky:
Dup with #70

PR #78, opened 2024-04-08
Added tvOS support

ErikEverson:
Added support to build tvOS targets so that consumers of [rustls-platform-verifier](https://github.com/rustls/rustls-platform-verifier) can use it with tvOS devices

Target support added for:
- x86_64-apple-tvos
- aarch64-apple-tvos
- aarch64-apple-tvos-sim

Motivation:
- We are adding support for tvOS to our library and consuming iOS application that relies on this Rust dependency directly. 

How was this tested:
- cargo test (passed)
- cargo +nightly build -Z build-std --lib --target x86_64-apple-tvos -v
- cargo +nightly build -Z build-std --lib --target aarch64-apple-tvos-v
- cargo +nightly build -Z build-std --lib --target aarch64-apple-tvos-sim -v

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,8 +16,25 @@ env:
   RUSTFLAGS: -D warnings -F unused_must_use
 
 jobs:
+  clippy-build-std:
+    name: Clippy (-Zbuild-std)
+    runs-on: macos-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          persist-credentials: false
+
+      - uses: dtolnay/rust-toolchain@nightly
+        with:
+          components: clippy
+
+      - name: Clippy (tvOS)
+        run: |
+          rustup component add rust-src --toolchain nightly-x86_64-apple-darwin
+          cargo +nightly clippy -Zbuild-std --target aarch64-apple-tvos
+
   clippy:
-    name: Clippy
+    name: Clippy (stable)
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
--- a/rustls-platform-verifier/Cargo.toml
+++ b/rustls-platform-verifier/Cargo.toml
@@ -35,7 +35,7 @@ base64 = { version = "0.21", optional = true } # Only used when the `cert-loggin
 jni = { version = "0.19", default-features = false, optional = true } # Only used during doc generation
 once_cell = "1.9"
 
-[target.'cfg(all(unix, not(target_os = "android"), not(target_os = "macos"), not(target_os = "ios")))'.dependencies]
+[target.'cfg(all(unix, not(target_os = "android"), not(target_os = "macos"), not(target_os = "ios"), not(target_os = "tvos")))'.dependencies]
 rustls-native-certs = "0.7"
 webpki = { package = "rustls-webpki", version = "0.102", features = ["ring", "alloc", "std"] }
 
@@ -52,11 +52,11 @@ webpki-roots = "0.26"
 [target.'cfg(target_os = "freebsd")'.dev-dependencies]
 webpki-roots = "0.26"
 
-[target.'cfg(any(target_os = "macos", target_os = "ios"))'.dependencies]
+[target.'cfg(any(target_os = "macos", target_os = "ios", target_os = "tvos"))'.dependencies]
 core-foundation = "0.9"
 core-foundation-sys = "0.8"
-security-framework = { version = "2.6", features = ["OSX_10_14"] }
-security-framework-sys = { version = "2.4", features = ["OSX_10_14"] }
+security-framework = { version = "2.10", features = ["OSX_10_14"] }
+security-framework-sys = { version = "2.10", features = ["OSX_10_14"] }
 
 [target.'cfg(windows)'.dependencies]
 winapi = { version = "0.3", features = ["wincrypt", "winerror"] }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -361,9 +361,9 @@ dependencies = [
 
 [[package]]
 name = "security-framework"
-version = "2.9.2"
+version = "2.10.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "05b64fb303737d99b81884b2c63433e9ae28abebe5eb5045dcdd175dc2ecf4de"
+checksum = "770452e37cad93e0a50d5abc3990d2bc351c36d0328f86cefec2f2fb206eaef6"
 dependencies = [
  "bitflags",
  "core-foundation",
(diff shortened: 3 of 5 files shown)


djc (approved):
Nice, thanks! Can you clean up the formatting issue?

complexspaces:
Is it possible to add a CI job that at least runs our Clippy checks on the tvOS target? It would be nice if this didn't regress quietly by mistake.

rustls-platform-verifier/Cargo.toml
```diff
@@ -52,11 +52,11 @@ webpki-roots = "0.26"
...
 webpki-roots = "0.26"
 
-[target.'cfg(any(target_os = "macos", target_os = "ios"))'.dependencies]
+[target.'cfg(any(target_os = "macos", target_os = "ios", target_os = "tvos"))'.dependencies]
 core-foundation = "0.9"
 core-foundation-sys = "0.8"
-security-framework = { version = "2.6", features = ["OSX_10_14"] }
-security-framework-sys = { version = "2.4", features = ["OSX_10_14"] }
+security-framework = { version = "2.10", features = ["OSX_10_14"] }
+security-framework-sys = { version = "2.10", features = ["OSX_10_14"] }
```
cpu, line 59: The upstream is a bit messy here :-/ No GitHub releases or changelog. Cargo.toml version updates are rolled in to other commits, etc.

I'm assuming this is for https://github.com/kornelski/rust-security-framework/pull/190?

ErikEverson: Yes the version bump is for those updates.

Otherwise I get the follwing import compile errors
`nresolved imports 'security_framework::certificate', 'security_framework::policy', 'security_framework::secure_transport', 'security_framework::trust'`

cpu: For the curious:

* [diff.rs for -sys 2.4.0 -> 2.1.0](https://diff.rs/security-framework-sys/2.4.0/2.10.0/Cargo.toml)
* [diff.rs for security-framework 2.6.0 -> 2.10.0](https://diff.rs/security-framework/2.6.0/2.10.0/Cargo.toml)

complexspaces: TIL about `diff.rs`, that's a very useful website.

cpu:
I think it's part of why this PR is marked a draft, but just to be explicit it would also be great if you could tidy up the commit history when you add the CI coverage. Thank you!

ErikEverson:
@complexspaces  I am not sure if Clippy can easily be run or at all on tier 3 targets. https://github.com/rust-lang/rust-clippy/issues/4573#issuecomment-535521420
Also I looked into running just Check. While possible it is quite involved and was only potentially solved two days ago here.
https://github.com/rust-lang/rust/issues/101172#issuecomment-1464146040

complexspaces:
@ErikEverson Thanks for taking an initial look at that! I checked out this branch locally, and I think that I was able to get Clippy to work on a recent nightly. Could you see if this is reproducible for you as well? I'm on an M1 macOS 14 system.

These are my versions:
```
cargo +nightly --version
cargo 1.79.0-nightly (0637083df 2024-04-02)

rustc +nightly --version
rustc 1.79.0-nightly (385fa9d84 2024-04-04)
```

Running `cargo clean` and then `cargo +nightly clippy -Zbuild-std --target aarch64-apple-tvos` completes in a few seconds after building the standard library. I tested its functionality by adding `#![warn(clippy::as_conversions)]` to `lib.rs` and then removing the `#[allow(clippy::as_conversions)]` annotation on `system_time_to_cfdate`. After running it again the warnings I'd expect appeared:
```
warning: using a potentially dangerous silent `as` conversion
  --> rustls-platform-verifier/src/verification/apple.rs:32:36
   |
32 |     let unix_adjustment = unsafe { kCFAbsoluteTimeIntervalSince1970 as u64 };
   |                                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = help: consider using a safe wrapper for this conversion
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#as_conversions
note: the lint level is defined here
  --> rustls-platform-verifier/src/lib.rs:4:9
   |
4  | #![warn(clippy::as_conversions)]
   |         ^^^^^^^^^^^^^^^^^^^^^^
   ```

If `-Zbuild-std` is placed _before_ `clippy` in the command it fails to work though.

**Comment by ErikEverson** [comment]
@complexspaces Awesome I will see if I get the same on my side. I was putting `-Zbuild-std` first

**Comment by cpu** [comment]
> ErikEverson force-pushed the EE/tvOS_support branch from 80cd835 to e9dc1b9 

Thanks! Commit history is looking better. A couple suggestions:

* I think rather than take the `git` dependency temporarily and then bumping the sec-framework dep versions separately it'd be best to do that all in one commit.
* Similarly, the fmt fix should get folded into commit that introduced the change being reformatted.

So the end state would be something like:

* One commit updating the required deps, adding the new cfg gates, and correctly formatted with `cargo fmt`.
* One commit updating CI.

**Comment by ErikEverson** [comment]
@cpu @complexspaces  I got clippy to work locally and on CI. One quick question should I run clippy against all tvOS targets or just one like iOS currently is?
tvOS targets:

- aarch64-apple-tvos
- aarch64-apple-tvos-sim
- x86_64-apple-tvos

iOS targets:

- aarch64-apple-ios
- aarch64-apple-ios-sim
- x86_64-apple-ios (only one with clippy ran)

**Comment by cpu** [comment]
>  One quick question should I run clippy against all tvOS targets or just one like iOS currently is?

I don't feel strongly. Matching the existing iOS target seems sensible to me.

**Comment by complexspaces** [comment]
I also don't think this matters too much, so let's just target the one like iOS does. We don't have any architecture or environment specific code in `rustls-platform-verifier` so it shouldn't matter.

File: .github/workflows/ci.yml [hunk 2]
```diff
@@ -16,8 +16,30 @@ env:
   RUSTFLAGS: -D warnings -F unused_must_use
 
 jobs:
+  check:
+    name: Clippy (nightly)
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os:
+          - macos-latest
```
cpu, line 25: I think it would be clearer if this had the matrix removed and replaced with `runs-on: macos-latest`. It only has 1 variant and we have to check for it again on L36 for the `if` guard. Hardcoding it to `runs-on: macos-latest` and dropping the guard seems easier.

Similarly I don't think it's fair to call this "Clippy (nightly)" which might give the impression it's nightly clippy for all targets when it's only testing MacOS + aarch64-apple-tvos. WDYT about naming it "Clippy (tvOS)"?

cpu (approved):
Thank you!

Looks good to me, I only had some nits about the CI task to suggest.

.github/workflows/ci.yml
ErikEverson: I meanly left it open for when other tier 3 target support is added like visionOS or watchOS targets.
Maybe name it "tier 3"?

@phatblat will likely be along at some point to add visionOS support here after rust core supports it.

ErikEverson: Yeah I will drop the matrix part.

ErikEverson: I removed the matrix and updated the naming to have Zbuild-std as that is the real reason that nightly is required for the tier 3 targets.

cpu: Great, thank you!

cpu (approved):

complexspaces (approved):

cpu:
I will start preparing a point release :rocket:

PR #80, opened 2024-04-08
docs: fix broken docs link to with_custom_certificate_verifier

jbr:
Would y'all be interested in a little example after this link that demonstrates providing a custom crypto provider?


djc (approved):

cpu (approved):
Thank you

Issue #79, opened 2024-04-08
Enable non default crypto config for initialization of Verifier


nnmkhang:
in `src/verification/windows.rs` there is a check to ensure that the default `CryptoProvider` must be used.

Enabling custom crypto providers will allow users who want to opt out of using the `rustls` the ability to configure platform verifier with their own crypto provider. 

From 'windows.rs` on line [427](https://github.com/rustls/rustls-platform-verifier/blob/main/rustls-platform-verifier/src/verification/windows.rs#L429)

```
impl Verifier {
    /// Creates a new instance of a TLS certificate verifier that utilizes the
    /// Windows certificate facilities. The rustls default [`CryptoProvider`]
    /// must be set before the verifier can be used.
    pub fn new() -> Self {
        Self {
            #[cfg(any(test, feature = "ffi-testing", feature = "dbg"))]
            test_only_root_ca_override: None,
            default_provider: OnceCell::new(),
        }
    }
````

and also line [448](https://github.com/rustls/rustls-platform-verifier/blob/main/rustls-platform-verifier/src/verification/windows.rs#L448C1-L456C6)

```
    fn get_provider(&self) -> &CryptoProvider {
        self.default_provider
            .get_or_init(|| {
                rustls::crypto::CryptoProvider::get_default()
                    .expect("rustls default CryptoProvider not set")
                    .clone()
            })
            .as_ref()
    }
```

My suggestion: allow passing in a `CryptoProvider` to the `Verifier`, and default to the `rustls` default if no `CryptoProvider` is supplied.

cpu:
> My suggestion: allow passing in a CryptoProvider to the Verifier, and default to the rustls default if no CryptoProvider is supplied.

This sounds like a good suggestion. WDYT about https://github.com/rustls/rustls-platform-verifier/pull/81  ?

cpu:
This was resolved by #81. Thanks again for filing the issue! I'm working on a release with this feature in https://github.com/rustls/rustls-platform-verifier/pull/88.

PR #81, opened 2024-04-09
feat: add Verifier::set_provider and Verifier::with_provider

jbr:
This is offered as a possible solution for #79, which I just encountered.

Implementation notes, let me know if any of these assumptions were wrong:

* I renamed `default_provider` fields to `crypto_provider` to make clear that it's not necessarily the process-default.
* I added both a `set_provider` and `with_provider`, the latter of which is convenient for `Verifier::new().with_provider(Arc::new(...))` but I'll replace with `Verifier::new_with_provider(...)` if that's more in line with this crate's style.
* ~I opted to panic if the crypto provider has already been set instead of returning an indication of failure, with the expectation that it's a bug if this is called multiple times on a given `Verifier`.~ We have `&mut`, no panic or error needed

### DIFF
--- a/rustls-platform-verifier/src/verification/android.rs
+++ b/rustls-platform-verifier/src/verification/android.rs
@@ -47,7 +47,7 @@ pub struct Verifier {
     /// Testing only: The root CA certificate to trust.
     #[cfg(any(test, feature = "ffi-testing"))]
     test_only_root_ca_override: Option<Vec<u8>>,
-    default_provider: OnceCell<Arc<CryptoProvider>>,
+    pub(super) crypto_provider: OnceCell<Arc<CryptoProvider>>,
 }
 
 impl Default for Verifier {
@@ -71,13 +71,16 @@ impl Drop for Verifier {
 
 impl Verifier {
     /// Creates a new instance of a TLS certificate verifier that utilizes the
-    /// Android certificate facilities. The rustls default [`CryptoProvider`]
-    /// must be set before the verifier can be used.
+    /// Android certificate facilities.
+    ///
+    /// A [`CryptoProvider`] must be set with
+    /// [`set_provider`][Verifier::set_provider]/[`with_provider`][Verifier::with_provider] or
+    /// [`CryptoProvider::install_default`] before the verifier can be used.
     pub fn new() -> Self {
         Self {
             #[cfg(any(test, feature = "ffi-testing"))]
             test_only_root_ca_override: None,
-            default_provider: OnceCell::new(),
+            crypto_provider: OnceCell::new(),
         }
     }
 
@@ -86,18 +89,8 @@ impl Verifier {
     pub(crate) fn new_with_fake_root(root: &[u8]) -> Self {
         Self {
             test_only_root_ca_override: Some(root.into()),
-            default_provider: OnceCell::new(),
+            crypto_provider: OnceCell::new(),
         }
-    }
-
-    fn get_provider(&self) -> &CryptoProvider {
-        self.default_provider
-            .get_or_init(|| {
-                rustls::crypto::CryptoProvider::get_default()
-                    .expect("rustls default CryptoProvider not set")
-                    .clone()
-            })
-            .as_ref()
     }
 
     fn verify_certificate(
--- a/rustls-platform-verifier/src/verification/mod.rs
+++ b/rustls-platform-verifier/src/verification/mod.rs
@@ -1,3 +1,6 @@
+use rustls::crypto::CryptoProvider;
+use std::sync::Arc;
+
 #[cfg(all(
     any(unix, target_arch = "wasm32"),
     not(target_os = "android"),
@@ -64,7 +67,7 @@ fn log_server_cert(_end_entity: &rustls::pki_types::CertificateDer<'_>) {
 #[cfg(any(windows, target_os = "macos", target_os = "ios"))]
 fn invalid_certificate(reason: impl Into<String>) -> rustls::Error {
     rustls::Error::InvalidCertificate(rustls::CertificateError::Other(rustls::OtherError(
-        std::sync::Arc::from(Box::from(reason.into())),
+        Arc::from(Box::from(reason.into())),
     )))
 }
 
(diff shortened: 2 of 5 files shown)


rustls-platform-verifier/src/verification/android.rs
```diff
@@ -77,21 +77,42 @@ impl Verifier {
...
         }
     }
 
+    /// Attempts to set the [`CryptoProvider`] for this `Verifier`.
+    ///
+    /// # Panics
+    ///
+    /// This function will panic if this verifier's crypto provider has already been set.
+    pub fn set_provider(&mut self, crypto_provider: Arc<CryptoProvider>) {
+        self.crypto_provider
```
djc, line 90: Why not propagate the error instead, letting the caller chose how to handle it?

cpu: I agree

cpu:
Thank you for picking this up! The general approach seems like the right solution.

rustls-platform-verifier/src/verification/android.rs
```diff
@@ -77,21 +77,42 @@ impl Verifier {
...
+            .set(crypto_provider)
+            .expect("crypto provider was already set");
+    }
+
+    /// Chainable constructor to set the [`CryptoProvider`] on a new [`Verifier`]
+    ///
+    /// # Panics
+    ///
+    /// This function will panic if this verifier's crypto provider has already been set.
+    pub fn with_provider(mut self, crypto_provider: Arc<CryptoProvider>) -> Self {
```
complexspaces, line 100: I wonder if it would be better/possible to raise the `set_provider` and `with_provider` methods (and their `impl` block) into the `verifier/mod.rs` level. This would help cut down on the code and comment duplication. As the platform-specific modules don't need to call these, we wouldn't end up with any circular code either.

As an example of what I mean:
```rust
<snip>

#[cfg(windows)]
mod windows;

#[cfg(windows)]
pub use windows::Verifier;

impl Verifier {
    pub fn set_provider(&mut self) { ... }
    pub fn with_provider(mut self) { ... }
}

<snip>
```

rustls-platform-verifier/src/verification/android.rs
jbr: I realized we have `&mut` here so it's infallible :tada:

rustls-platform-verifier/src/verification/android.rs
jbr: Will do, I like that but figured there was a reason there wasn't already a shared-behavior block. We'll need to omit `feature = "other"` which doesn't have a crypto provider, but that isn't much complexity

rustls-platform-verifier/src/verification/mod.rs
```diff
@@ -77,3 +77,29 @@ fn invalid_certificate(reason: impl Into<String>) -> rustls::Error {
 /// - id-kp-serverAuth
 // TODO: Chromium also allows for `OID_ANY_EKU` on Android.
 pub const ALLOWED_EKUS: &[&str] = &["1.3.6.1.5.5.7.3.1"];
+
+#[cfg(any(target_os = "macos", target_os = "ios", target_os = "android", windows))]
+impl Verifier {
```
cpu, line 82: Is it possible to extend support for supplying the crypto provider to unix/linux as well?

I think it's largely a matter of wiring up `WebPkiServerVerifier::builder_with_provider()` where we're using `WebPkiServerVerifier::builder()`: https://github.com/rustls/rustls-platform-verifier/blob/b6f6334dc9d56e53d4a722ab801d76b935250e04/rustls-platform-verifier/src/verification/others.rs#L88-L90

nnmkhang:
Hi Everyone!

Thanks @jbr for picking this up so fast, I'm not sure what type of testing you have done but I tried to verify this on my end with a custom crypto provider but I am running into some issues. 

Here is the snippet of code that I am trying to get working.
```
let mut config = rustls::ClientConfig::builder_with_provider(Arc::new(default_symcrypt_provider()))
        .with_safe_default_protocol_versions()
        .unwrap()
        .dangerous()
        .with_custom_certificate_verifier(Arc::new(Verifier::set_provider(Arc::new(default_symcrypt_provider))))
        .with_no_client_auth();
```
And im getting the following error, what am I doing wrong here?

```
error[E0277]: the trait bound `(): ServerCertVerifier` is not satisfied
  --> bin/sample_internet_client_platform.rs:20:43
   |
20 |         .with_custom_certificate_verifier(Arc::new(Verifier::set_provider(Arc::new(default_symcrypt_provider))))
   |                                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `ServerCertVerifier` is not implemented for `()`
   |
   = help: the following other types implement trait `ServerCertVerifier`:
             Verifier
             WebPkiServerVerifier
   = note: required for the cast from `Arc<()>` to `Arc<(dyn ServerCertVerifier + 'static)>`
```

I've also noticed on the `rustls` [documentation]([WantsVerifier](https://docs.rs/rustls/latest/rustls/struct.WantsVerifier.html)) for `ClientConfig` that `builder_with_provider` does not take in the `WantsVerifier` struct.

Please forgive my new-ness to rust but would this be the cause of the error? Or is there something wrong with the configuration on my end?

Appreciate the help here!

jbr:
@nnmkhang It's still a PR and CI is currently failing, but with the current iteration, I think you want code that looks like

```rust
ClientConfig::builder_with_provider(provider.clone())
   .with_safe_default_protocol_versions()
   .unwrap()
   .dangerous()
   .with_custom_certificate_verifier(Arc::new(Verifier::new().with_provider(provider)))
   .with_no_client_auth()
```

Note the `Verifier::new().with_provider(provider)` instead of `Verifier::set_provider(provider)`

rustls-platform-verifier/src/verification/mod.rs
```diff
@@ -77,3 +77,36 @@ fn invalid_certificate(reason: impl Into<String>) -> rustls::Error {
...
 pub const ALLOWED_EKUS: &[&str] = &["1.3.6.1.5.5.7.3.1"];
+
+impl Verifier {
+    /// Chainable setter to configure the [`CryptoProvider`][rustls::crypto::CryptoProvider] for this `Verifier`.
+    ///
+    /// This will be used instead of the rustls processs-default `CryptoProvider`, even if one has
+    /// been installed.
+    pub fn with_provider(
+        mut self,
+        crypto_provider: std::sync::Arc<rustls::crypto::CryptoProvider>,
```
complexspaces, line 88: Nit: Let's import `Arc` instead of fully qualifying it three times in this block.

rustls-platform-verifier/src/verification/others.rs
```diff
@@ -28,17 +29,25 @@ pub struct Verifier {
...
 }
 
 impl Verifier {
     /// Creates a new verifier whose certificate validation is provided by
     /// WebPKI, using root certificates provided by the platform.
+    ///
+    /// A [`CryptoProvider`] must be set with
+    /// [`set_provider`][Verifier::set_provider]/[`with_provider`][Verifier::with_provider] or
+    /// [`CryptoProvider::install_default`] before the verifier can be used.
+
```
complexspaces, line 43: Nit: Please remove this extra newline.

complexspaces (approved):
Thanks for working through this with us! I think this is looking quite nice now.

cpu (approved):
Really nice! Thanks again.

cpu (approved):

nnmkhang:
@jbr, Just tried it again on my my machine and it works with the custom provider and the platform verifier enabled! Verified with event viewer on windows as well.

Thanks for your help and sorry for my silly mistake.

cpu:
I think we should square away #86 and #78 and then consider a 0.3.1 point release.

complexspaces:
Sounds good to me.

cpu:
[v/0.3.1](https://github.com/rustls/rustls-platform-verifier/releases/tag/v%2F0.3.1) is now available with this change included.

PR #82, opened 2024-04-09
Vet docs in CI

cpu:
Quick follow-up to https://github.com/rustls/rustls-platform-verifier/pull/80

It looks like this repo wasn't building documentation in CI. This branch adds a step for that, forbidding warnings. Additionally fixes a warning in test-only rustdoc that is raised when using `--document-private-items`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -229,3 +229,20 @@ jobs:
       
       - name: Verify release artifact
         run: ./ci/verify_android_release.sh
+
+  docs:
+    name: Check for documentation errors
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout sources
+        uses: actions/checkout@v3
+        with:
+          persist-credentials: false
+
+      - name: Install rust toolchain
+        uses: dtolnay/rust-toolchain@stable
+
+      - name: Build documentation
+        run: cargo doc --locked --all-features --no-deps --document-private-items
+        env:
+          RUSTDOCFLAGS: -Dwarnings
--- a/rustls-platform-verifier/src/tests/mod.rs
+++ b/rustls-platform-verifier/src/tests/mod.rs
@@ -52,7 +52,7 @@ pub fn assert_cert_error_eq<E: StdError + PartialEq + 'static>(
     }
 }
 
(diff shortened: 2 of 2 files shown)


ctz (approved):

.github/workflows/ci.yml
```diff
@@ -229,3 +229,20 @@ jobs:
...
+    name: Check for documentation errors
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout sources
+        uses: actions/checkout@v3
+        with:
+          persist-credentials: false
+
+      - name: Install rust toolchain
+        uses: dtolnay/rust-toolchain@nightly
```
complexspaces, line 243: Question: Is using `nightly` intentional here when most of CI uses stable?

cpu: I mainly took it for consistency with the rustls and webpki repos. 

We do that in those repos to be able to [use `doc_auto_cfg`](https://github.com/rustls/rustls/blob/fa605bfff02d7c2b7b31cca01ea2f5faf003fc3d/rustls/src/lib.rs#L337-L338). Probably we should use that here too. I will adjust.

cpu: Add `doc_auto_cfg` in a separate commit.

rustls-platform-verifier/src/lib.rs
```diff
@@ -1,4 +1,4 @@
-#![cfg_attr(docsrs, feature(doc_cfg))]
+#![cfg_attr(docsrs, feature(doc_cfg, doc_auto_cfg))]
```
complexspaces, line 1: Are we sure this isn't going to break again? This was [removed before](https://github.com/rustls/rustls-platform-verifier/pull/57/files#diff-8660cd1844f655870a5550ff30da2ac758384062af24800f5296599e69b95f27R1) in a previous PR of mine because it caused the wrong feature gates to appear on `docs.rs`.

cpu: Forgot about that :+1: 

I backed out the `doc_auto_cfg` change and switched to using stable rust for the documentation task.

complexspaces (approved):

PR #83, opened 2024-04-09
ci: add dependabot, update actions

cpu:
In https://github.com/rustls/rustls-platform-verifier/pull/82 I noticed the GitHub `actions/checkout` version was inconsistent.

This branch fixes that, updates the other actions that were out of date, and adds a similar dependabot config as we use in the base Rustls repo to help keep these up to date & consistent moving forward.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,15 @@
+version: 2
+updates:
+- package-ecosystem: cargo
+  directory: "/"
+  schedule:
+    interval: weekly
+  open-pull-requests-limit: 5
+  groups:
+    crates-io:
+      patterns:
+        - "*"
+- package-ecosystem: github-actions
+  directory: "/"
+  schedule:
+    interval: weekly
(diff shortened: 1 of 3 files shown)


djc (approved):

.github/dependabot.yml
```diff
@@ -0,0 +1,15 @@
+version: 2
+updates:
+- package-ecosystem: cargo
+  directory: "/"
+  schedule:
+    interval: weekly
```
complexspaces, line 6: FWIW I'm not sold on the benefit/noise ratio of Dependabot, especially when manual work is required to get any _important_ upgrades like to `rustls` and other fundamental crates this repository relies on. This isn't a blocker though.

cpu: I think it's useful for keeping the lockfile up-to-date with semver compatible updates. Yes, semver incompatible updates will always require manual work.

complexspaces: I'm not sure we agree on the usefulness because our lockfile doesn't get distributed or matter to downstream consumers of the library, but my thoughts here aren't a blocker.

complexspaces (approved):

.github/dependabot.yml
cpu: Fair. We can back it out if it ends up being annoying.

PR #86, opened 2024-04-09
feat: rustls_platform_verifier::tls_config_with_provider

jbr:
The primary entry point for this crate `tls_config` is quite convenient, but opting out of that in order to configure a crypto provider adds a bit more complexity and opportunity for misconfiguration. This PR offers a solution to "do what `tls_config` does, but with a crypto provider".

based on #81

### DIFF
--- a/rustls-platform-verifier/src/lib.rs
+++ b/rustls-platform-verifier/src/lib.rs
@@ -65,6 +65,25 @@ pub fn tls_config() -> ClientConfig {
         .with_no_client_auth()
 }
 
+/// Attempts to construct a `rustls` configuration that verifies TLS certificates in the best way
+/// for the underlying OS platform, using the provided
+/// [`CryptoProvider`][rustls::crypto::CryptoProvider].
+///
+/// See [`tls_config`] for further documentation.
+///
+/// # Errors
+///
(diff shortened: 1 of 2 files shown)


cpu (approved):
Good idea, thank you.

Would you mind updating the [README installation and setup](https://github.com/rustls/rustls-platform-verifier?tab=readme-ov-file#installation-and-setup) section to briefly mention this alongside the existing `tls_config()` coverage?

rustls-platform-verifier/src/lib.rs
```diff
@@ -65,6 +65,23 @@ pub fn tls_config() -> ClientConfig {
         .with_no_client_auth()
 }
 
+/// Attempts to construct a `rustls` configuration that verifies TLS certificates in the best way
+/// for the underlying OS platform, using the provided
+/// [`CryptoProvider`][rustls::crypto::CryptoProvider].
+///
```
complexspaces, line 71: Suggestion: It might be nice to say something like "See `tls_config` for an example and other details" here since that also mentions how to configure other parts of `ClientConfig`.

cpu (approved):

complexspaces (approved):

cpu:
This was included in [v/0.3.1](https://github.com/rustls/rustls-platform-verifier/releases/tag/v%2F0.3.1).

PR #103, opened 2024-06-25
Minimise features demanded from rustls-webpki crate

### LINKED ISSUE #102: Mandatory dependency on ring on unix platforms (including Linux)
`rustls-platform-verifier/Cargo.toml` contains:
```toml
[target.'cfg(all(unix, not(target_os = "android"), not(target_os = "macos"), not(target_os = "ios"), not(target_os = "tvos")))'.dependencies]
rustls-native-certs = "0.7"
webpki = { package = "rustls-webpki", version = "0.102", features = ["ring", "alloc", "std"] }
```

As far as I can tell, this force-enables `ring` on any `unix` platform that isn't `android` or an apple platform.

I maintain a crate with an indirect dependency on `rustls-platform-verifier`, and I'd like to be able to offer my users the option of using `ring`, `aws-lc-rs`, or a custom crypto provider.

Would it be possible to remove this mandatory dependency?

ctz:
This dependency exists so this crate can access
`webpki::Error` and `webpki::Error::RequiredEkuNotFound`, which are unconditionally available.

fixes #102

### DIFF
--- a/rustls-platform-verifier/Cargo.toml
+++ b/rustls-platform-verifier/Cargo.toml
@@ -37,12 +37,12 @@ once_cell = "1.9"
 
 [target.'cfg(all(unix, not(target_os = "android"), not(target_os = "macos"), not(target_os = "ios"), not(target_os = "tvos")))'.dependencies]
 rustls-native-certs = "0.7"
-webpki = { package = "rustls-webpki", version = "0.102", features = ["ring", "alloc", "std"] }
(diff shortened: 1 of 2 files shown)


ctz:
> CI / Test (windows-latest) (pull_request) Failing after 1m

This is also happening on main; added a remedial commit.

djc (approved):

PR #105, opened 2024-06-26
Enable `rustls_native_certs::load_native_certs` on FreeBSD

### LINKED ISSUE #104: Possible FreeBSD certificate detection failure
Hello there! After the Rustup team's recent attempt to migrate to `rustls-platform-verifier` (https://github.com/rust-lang/rustup/pull/3903) we have noticed a CI regression as follows:

> error: could not download file from 'https://static.rust-lang.org/rustup/release-stable.toml' to '/tmp/rustup-update68roIr/release-stable.toml': failed to make network request: error sending request for url (https://static.rust-lang.org/rustup/release-stable.toml): client error (Connect): received fatal alert: DecodeError

> I'm not sure what this `DecodeError` is. Could it be `X.509`-related?
_https://github.com/rust-lang/rustup/issues/3908#issue-2372437430_

Interestingly enough, before the migration we have used `reqwest`'s integration of `rustls-native-certs v0.7.0` and it didn't fail:

> Before (d85502ca):
> 
> ```console
> > cargo tree -i webpki-roots --target=x86_64-unknown-freebsd
> error: package ID specification `webpki-roots` did not match any packages
> 
> > cargo tree -i rustls-native-certs --target=x86_64-unknown-freebsd
> rustls-native-certs v0.7.0
> └── reqwest v0.12.4
>     └── download v1.27.1
>         └── rustup v1.27.1
> ```
> 
> After (f48df22e):
> 
> ```console
> > cargo tree -i webpki-roots --target=x86_64-unknown-freebsd
> warning: nothing to print.
> 
> > cargo tree -i rustls-native-certs --target=x86_64-unknown-freebsd
> rustls-native-certs v0.7.0
> └── rustls-platform-verifier v0.3.1
>     └── download v1.27.1
>         └── rustup v1.27.1
> ```
_https://github.com/rust-lang/rustup/issues/3908#issuecomment-2191355535_

---

Comparing the usage of `rustls_native_certs::load_native_certs` in `reqwest` and `rustls-platform-verifier`, I have noticed that here this function is under a guard that is disabled on FreeBSD (https://github.com/rust-lang/rustup/issues/3908#issuecomment-2191368278):

https://github.com/rustls/rustls-platform-verifier/blob/2b3bfbe9789b15748364e5b8c1b7f93848dea26e/rustls-platform-verifier/src/verification/others.rs#L118-L119

Could this be a mistake of some sort? Does it have something to do with the snippet below?

https://github.com/rustls/rustls-platform-verifier/blob/2b3bfbe9789b15748364e5b8c1b7f93848dea26e/rustls-platform-verifier/Cargo.toml#L51-L53

Many thanks in advance!

rami3l:
Closes #104.


rustls-platform-verifier/src/verification/others.rs
```diff
@@ -115,7 +115,7 @@ impl Verifier {
             );
         }
 
-        #[cfg(all(target_os = "linux", not(target_arch = "wasm32")))]
+        #[cfg(all(unix, not(target_os = "android"), not(target_os = "macos"), not(target_os = "ios"), not(target_os = "tvos")))]
```
cpu, line 118: Is the removal of `not(target_arch = "wasm32")` intentional? 

I _think_ that `wasm32-unknown-emscripten` (as one example) is `arch == wasm32` and `target-family == [unix, wasm]` and would be included here even though there's no native certs to load for that target. Or am I overlooking something?

rami3l: @cpu To be frank I'm not sure, as the original request was:

> I guess this should be equivalent to the cfg expression that adds the dependency on rustls-native-certs in the first place.

Please feel free to adjust the guard as you see fit.

cpu: I think probably both the `Cargo.toml` dependency gating for `rustls-native-certs` and this should have a `not(target_arch = "wasm32")` and it was probably an oversight :thinking: 

Does that make sense to you @ctz?

ctz: Yes, that makes sense to me. There's no point in calling or building `rustls-native-certs` for wasm. (And if someone wants that to work for wasi we can deal with that when they ask.)

cpu: > Please feel free to adjust the guard as you see fit.

I'll push a revision in a moment.

cpu (approved):
LGTM, but since I've pushed commits I won't merge it myself.

ctz (approved):
The more I think about it I'm not sure the original cfg expression made sense (linux, but not on wasm32 -- eh?) This is better.

cpu:
Thanks @rami3l !

Issue #69, opened 2024-03-06
Self signed certificates with user-installed CA show as revoked on Android


dani-garcia:
I’ve been testing the use of this library and while it’s been working great so far on Mac, iOS and Android for certificates generated from system-trusted CAs, I’ve found that Android can’t seem to deal with certificates signed by user-installed CAs, and always reports them as Revoked.

My tests have gone like this:
- Make an HTTPS request to a site using a self signed certificate (for example `https://untrusted-root.badssl.com/`). This returns `InvalidCertificate(UnknownIssuer)` on all platforms, which is expected.
- Get the CA certificate from that domain and install it in the OS trust store
- Make a new request again, this succeeds in Mac and iOS, but on Android I get `InvalidCertificate(Revoked)`

I've also tested other certificates signed by a local CA in a dev server, like the ones generated by [mkcert](https://github.com/FiloSottile/mkcert), and I can reproduce the same problem there too.

I've also checked that the requests succeed if you make them with an Android native HTTP client (like Kotlins Ktor HTTP client), which suggests to me that the certificate is installed correctly at least.

Any ideas what could be causing this issue?

PS: I have made a small app to reproduce the issue if that helps, it's just using the `ureq` library with `rustls-platform-verifier`, and can be used from the included Android app and the CLI: https://github.com/dani-garcia/rpv-selfsigned

cpu:
Hi @dani-garcia,

Thanks for opening an issue and linking to your reproducer. This is a curious situation and I don't have any immediate ideas. 

Based on what you describe it sounds like `X509TrustManager.checkServerTrusted` is passing for the presented chain using the Android trust store, but `CertPathValidator.validate` is throwing an exception we map to the revoked state (possibly masking an underlying error of a different sort?).

I think reproducing this with a test run in the emulator would be a good next step. I think it's likely that either the revocation checking is expecting certain metadata that isn't present in the added trust anchor and is failing closed, or there is a separate validation error being presented incorrectly as a revocation error. The main open question in my mind is the best way to programmatically inject a new trust anchor into the emulator's system trust bundle.

cpu:
> The main open question in my mind is the best way to programmatically inject a new trust anchor into the emulator's system trust bundle

Perhaps like this: https://docs.mitmproxy.org/stable/howto-install-system-trusted-ca-android/

cpu:
> Perhaps like this: https://docs.mitmproxy.org/stable/howto-install-system-trusted-ca-android/

[...]

PR #108, opened 2024-07-04
Android: Don't attempt to check revocation on non-public certificates

complexspaces:
This PR fixes an issue on Android where the verifier was attempting to enforce revocation constraints even on self-signed certificates that don't (nor should need to) supply revocation information. This PR fixes this by bringing back our previous `isKnownRoot` check and using this to determine if we should even try the revocation codepaths. If a certificate isn't a known root, we don't let Android enforce revocation information.

A small cutout was left for cases where an explicit stapled OSCP response is provided by the server. This is for two reasons:
- Our test suite needs to be able to verify a mocked, frozen-in-time, OCSP response is confirmed as revoked.
- There _might_ be a case where someone has this setup in the real world and they are probably expecting the OSCP data to be checked by clients.

Closes https://github.com/rustls/rustls-platform-verifier/issues/69

### DIFF
--- a/android/rustls-platform-verifier/src/androidTest/java/org/rustls/platformverifier/CertificateVerifierTests.kt
+++ b/android/rustls-platform-verifier/src/androidTest/java/org/rustls/platformverifier/CertificateVerifierTests.kt
@@ -4,6 +4,7 @@ import android.content.Context
 import androidx.test.ext.junit.runners.AndroidJUnit4
 import androidx.test.platform.app.InstrumentationRegistry
 import org.junit.Assert.assertEquals
+import org.junit.Assert.assertTrue
 import org.junit.BeforeClass
 import org.junit.Test
 import org.junit.runner.RunWith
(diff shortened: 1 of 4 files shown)


android/rustls-platform-verifier/src/androidTest/java/org/rustls/platformverifier/CertificateVerifierTests.kt
```diff
@@ -50,4 +51,22 @@ class CertificateVerifierTests {
         val result = verifyMockRootUsage(context)
         assertEquals(FAILURE_MSG, SUCCESS_MARKER, result)
     }
+
+    // Note:
+    //
+    // - Full negative path (`CertificateVerifier`'s flow for unknown roots,
+    // are already exercised via `runMockTestSuite`.
```
cpu, line 58: small nit: missing a closing `)` paren for this line :-)

cpu (approved):
Thanks for picking this up! :stars: 

The fix seems reasonable to me given everything we know about the various limitations/constraints we're operating with here.

android/rustls-platform-verifier/src/androidTest/java/org/rustls/platformverifier/CertificateVerifierTests.kt
complexspaces: Thanks! This is fixed now.

PR #110, opened 2024-07-12
Ensure license files are included in published crates

decathorpe:
Both the Apache-2.0 and MIT licenses require that redistributed sources contain a copy of the original license text.

The fact that it is missing here but other rustls projects all do this correctly makes me think that this was simply an oversight when this new project was set up.

### DIFF
--- a/android-release-support/LICENSE-APACHE
+++ b/android-release-support/LICENSE-APACHE
@@ -1,0 +1,1 @@
+../LICENSE-APACHE
--- a/android-release-support/LICENSE-MIT
+++ b/android-release-support/LICENSE-MIT
@@ -1,0 +1,1 @@
+../LICENSE-MIT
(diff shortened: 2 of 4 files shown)


cpu (approved):
Thanks! I think this was an oversight when we split the crate into a workspace of sub crates.

cpu:
> CI / Test (Android) (pull_request) Failing after 3m

Looks like we need to refresh some vendored test data. Can be ignored for this branch.

djc (approved):

complexspaces (approved):
Thanks for catching this!

decathorpe:
Thanks! Would it be possible to get a 0.3.3 release published to crates.io with this change?

cpu:
> Would it be possible to get a 0.3.3 release published to crates.io with this change?

I can prepare this ~Monday if nobody beats me to it.

cpu:
> I can prepare this ~Monday

https://github.com/rustls/rustls-platform-verifier/pull/114

PR #111, opened 2024-07-13
tests: fix `update_valid_ee_certs.rs` script, update vendored test data

cpu:
The syntax for the Cargo script unstable cargo feature changed out from under us, producing syntax errors running `update_valid_ee_certs.rs` with recent nightly tooling. This branch updates the syntax to match the [documented example script](https://github.com/rust-lang/cargo/blob/master/src/doc/src/reference/unstable.md#script). Along the way I also updated the `reqwest` version used by the script to the latest. Notably this synchronizes on `rustls` versions and saves some build time.

With the tooling fixed we can update the stale vendored end-entity cert. test data & static verification timestamp. The LE issuance chain [has changed](https://letsencrypt.org/2024/04/12/changes-to-issuance-chains) and so the intermediate & root also need updating. It would be nice if the tooling supported this too, but for now I've done it manually using the `.der` provided in [Let's Encrypt docs](https://letsencrypt.org/certificates/).


complexspaces (approved):
Nice! This looks good to me. We've switched the LE chains from RSA to EC IIUC?

cpu:
>  We've switched the LE chains from RSA to EC IIUC?

The old end entity cert was also P-256 ECDSA, but it was issued ee -> R3 (retired) -> ISRG Root X1 (cross-signed by DST Root X3).

This branch has ee -> E6 -> ISRG Root X2 (cross-signed by ISRG Root X1). Notably LE plans to switch between using E6 and E5 per-issuance for ECDSA subjects, so that's why it would be nice to have our tooling update the intermediate in addition to the leaf; we can expect to need to refresh the intermediate ~50% of the time we refresh the end entity cert going forward.

Issue #117, opened 2024-07-30
Test failures on `main`


cpu:
Noticed CI is failing [the Windows latest](https://github.com/rustls/rustls-platform-verifier/actions/runs/10167528562/job/28120139624) job:

```
---- tests::verification_mock::tests::stapled_revoked_dns stdout ----
thread 'tests::verification_mock::tests::stapled_revoked_dns' panicked at rustls-platform-verifier\src\tests\mod.rs:51:9:
assertion `left == right` failed
  left: Ok(())
 right: Err(InvalidCertificate(Revoked))
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

---- tests::verification_mock::tests::stapled_revoked_ipv4 stdout ----
thread 'tests::verification_mock::tests::stapled_revoked_ipv4' panicked at rustls-platform-verifier\src\tests\mod.rs:51:9:
assertion `left == right` failed
  left: Ok(())
 right: Err(InvalidCertificate(Revoked))

---- tests::verification_mock::tests::stapled_revoked_ipv6 stdout ----
thread 'tests::verification_mock::tests::stapled_revoked_ipv6' panicked at rustls-platform-verifier\src\tests\mod.rs:51:9:
assertion `left == right` failed
  left: Ok(())
 right: Err(InvalidCertificate(Revoked))


failures:
    tests::verification_mock::tests::stapled_revoked_dns
    tests::verification_mock::tests::stapled_revoked_ipv4
    tests::verification_mock::tests::stapled_revoked_ipv6
```

And the [Android task](https://github.com/rustls/rustls-platform-verifier/actions/runs/10167528562/job/28120138120):

```
07-30 18:14:51.040  3199  3229 I rustls_platform_verif..: verifying Err(InvalidCertificate(UnknownIssuer))
07-30 18:14:51.042  3199  3229 W rustls_platform_verif..: certificate was not trusted: java.
org.rustls.platformverifier.CertificateVerifierTests > runMockTestSuite[test(AVD) - 9] FAILED 
	org.junit.ComparisonFailure: A test failed. Check the logs above for Rust panics. expected:<[success]> but was:<[test failed!]>
	at org.junit.Assert.assertEquals(Assert.java:115)
security.cert.CertPathValidatorException: Trust anchor for certification path not found.
```

I won't have time to look into this right away, so filing for a rainy day.

complexspaces:
I think that the mock suite's root expired the other day:
[...]

PR #119, opened 2024-08-05
Update deny config & bump base64 to 0.22

djc:
Unblock: 

- #118

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -42,6 +42,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "414dcefbc63d77c526a76b3afcf6fbb9b5e2791c19c3aa2297733208750c6e53"
 
 [[package]]
+name = "base64"
+version = "0.22.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "72b3254f16251a8381aa12e40e3c4d2f0199f8c6508fbecb9d91f575e0fbb8c6"
+
+[[package]]
 name = "bitflags"
 version = "1.3.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -295,7 +301,7 @@ version = "2.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "35e4980fa29e4c4b212ffb3db068a564cbf560e51d3944b7c88bd8bf5bec64f4"
 dependencies = [
- "base64",
+ "base64 0.21.3",
  "rustls-pki-types",
 ]
 
@@ -310,7 +316,7 @@ name = "rustls-platform-verifier"
 version = "0.3.3"
 dependencies = [
  "android_logger",
- "base64",
+ "base64 0.22.1",
  "core-foundation",
  "core-foundation-sys",
  "jni",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,9 +4,9 @@ version = 3
 
 [[package]]
 name = "aho-corasick"
-version = "1.0.5"
+version = "1.1.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "0c378d78423fdad8089616f827526ee33c19f2fddbd5de1629152c9593ba4783"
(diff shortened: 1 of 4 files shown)


ctz (approved):

djc:
Not sure what's going on with the Windows CI jobs here.

cpu (approved):

cpu:
> Not sure what's going on with the Windows CI jobs here.

I'm not sure either, but it's unrelated to this branch. I opened https://github.com/rustls/rustls-platform-verifier/issues/117 for the failures the other day but haven't investigated yet.

Cargo.lock
```diff
@@ -41,6 +41,12 @@ version = "0.21.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "414dcefbc63d77c526a76b3afcf6fbb9b5e2791c19c3aa2297733208750c6e53"
 
+[[package]]
+name = "base64"
+version = "0.22.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "72b3254f16251a8381aa12e40e3c4d2f0199f8c6508fbecb9d91f575e0fbb8c6"
+
```
complexspaces, line 49: Question: Why are we bumping base64 when it just duplicates it in-tree compared to the rest of the lower-level `rustls` crates?

djc: It has been updated in the upstream crates, but we haven't been merging `Cargo.lock` updates as proposed by Dependabot in this repo.

cpu: > we haven't been merging Cargo.lock updates as proposed by Dependabot in this repo.

In retrospect I think turning that on for this repo was probably a mistake. I'd be OK with removing it again.

djc: Why do you think so?

cpu: > Why do you think so?

The update PRs sit without reviews/merge and so the bot updates just become churn notifications.

From my perspective the norms for reviewing those PRs hasn't been set in this repo, or written down explicitly. I personally don't feel enough ownership here to blanket +1 updates that don't affect MSRV like I might in `rustls/rustls` or `rustls/webpki`.

I remember @complexspaces pushing back on landing the automation and in retrospect I think it would have been better to work through some kind of policy/expectation-setting before landing the automation. Lesson learned :-)

cpu: > I'd be OK with removing it again

Here's a PR for that: https://github.com/rustls/rustls-platform-verifier/pull/120

If nothing else it gives us a place to continue this discussion separate from the `cargo deny` issue.

complexspaces: I hope no one feels bad about the current state of the lockfile/dependency updates, that wasn't my intention at all. I tend to not look at the PRs because my OSS time is already fairly limited.

I did push back when first proposed because it seemed weird to me to update dependencies via lockfile just because they had updated, since this is generally not what I do at work or personal projects. I tend to only do dependency upgrades when there's a new direct dependency and then make sure the lockfile is clean as an end result. 

This is definitely something I'd be happy discussing more though, since I don't think there's one correct way to do this across all projects.

cpu: > I hope no one feels bad about the current state of the lockfile/dependency updates

No hard feelings here!

djc: Nor here!

djc: @complexspaces I've tacked on a separate commit here which just did `cargo update`, which shows that this crate only depends on base64 0.22 after these changes.

djc:
So I guess one potential benefit of keeping `Cargo.lock` up to date is that we will notice sooner when upstream dependencies adopt a newer MSRV, which might affect our MSRV (although this is potentially not such a big deal for semver-compatible updates).

cpu:
@djc Can you rebase this on `main` for the test data fixes? 

@complexspaces Anything else you want to see done with this branch? It'd be nice to have passing CI for community contributions like https://github.com/rustls/rustls-platform-verifier/pull/122

djc:
Rebased!

cpu (approved):

complexspaces (approved):

complexspaces:
> Anything else you want to see done with this branch?

Nope, things look good now 👍. Sorry, it was a busy week.

PR #120, opened 2024-08-05
ci: disable dependabot cargo updates

cpu:
Until we have a better sense of how to handle the update PRs it produces let's turn this off for now. I think I was premature in enabling it. Inspired by [discussion in this PR thread](https://github.com/rustls/rustls-platform-verifier/pull/119#discussion_r1704544595).

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,14 +1,5 @@
 version: 2
 updates:
-- package-ecosystem: cargo
-  directory: "/"
-  schedule:
-    interval: weekly
-  open-pull-requests-limit: 5
-  groups:
-    crates-io:
-      patterns:
-        - "*"
 - package-ecosystem: github-actions
   directory: "/"
   schedule:


complexspaces:
> Until we have a better sense of how to handle the update PRs it produces let's turn this off for now.

Did you want to discuss that in this PR or would it be better to open a followup issue where the topic can happen? I don't mind either way.

cpu:
> Did you want to discuss that in this PR

WFM!

I think there's two topics:

1. whether it's worth maintaining `Cargo.lock` updates for compatible updates. We do it in the other Rustls repos, but I don't think I can articulate a very meaningful argument for why we should do it here.
2. how they should be reviewed. In the other Rustls repos we don't (at least as I understand things) do more than a cursory review of the update. When we were using Dependabot I would review the updates in https://diff.rs and pay particular attention to things like `build.rs` updates, but for large deps like `aws-lc-rs`/`aws-lc-sys` I certainly don't review every changed line. I also haven't been keeping up with that practice since switching to Renovate. How are other folks handling these?

djc:
I generally scroll through the `Cargo.lock` changes looking for any new dependencies, but I don't look at the actual changes made in semver-compatible upstream versions.

djc (commented):
Should we also disable this for `github-actions`?

djc:
A potential third topic for discussion: being more explicit about ownership. So far I think we've mostly deferred to @complexspaces on any half-way nuanced changes, but the downside is that @complexspaces has pretty limited bandwidth certainly compared to the other rustls org maintainers. Maybe we can get to a tweaked proposed governance model?

cpu:
> Should we also disable this for github-actions?

These seem infrequent enough that I was OK leaving it on, but happy to revisit.

worldwise001:
Hey folks, @complexspaces directly reports to me currently (I'm director of seceng at 1Password); I don't feel particularly strongly one way or another as to what the governance model of this library/project should look like, but I am open to discussion as to how/where we can adjust or increase 1Password involvement/support. :)

djc:
> Hey folks, @complexspaces directly reports to me currently (I'm director of seceng at 1Password); I don't feel particularly strongly one way or another as to what the governance model of this library/project should look like, but I am open to discussion as to how/where we can adjust or increase 1Password involvement/support. :)
> 
> Let's bring this discussion into another issue?

Opened https://github.com/rustls/rustls-platform-verifier/issues/125 for this.

djc (approved):

PR #121, opened 2024-08-05
tests: regenerate verification_mock data

cpu:
Commits the result of running `src/tests/verification_mock/ca.go`.

The `root1.crt` CA and intermediate expired and were breaking tests in CI ([example](https://github.com/rustls/rustls-platform-verifier/actions/runs/10253848961)).

A larger mystery remains w.r.t why the failure observed on Windows for stapled OCSP revocation doesn't match our expectation given a root cause of an expired trust anchor/intermediate...

Relates to https://github.com/rustls/rustls-platform-verifier/issues/117


djc (approved):

complexspaces (approved):
Thanks for the immediate fixes!

I edited the PR description to not close the attached issue since we both agree this warrants some more investigation on the Windows side before closing 😬.

PR #127, opened 2024-08-23
Fix UB in Windows verifier EKU handling 

### LINKED ISSUE #126: TLS Runtime Issues for GitHub Artifacts built on Windows
Hello 😊 We have discovered a weird runtime TLS bug happening for artifacts built on the GitHub Windows runners. We have an example repository showcasing a simplified example here:

https://github.com/Thomas-Avery/test-rustls-platform-verifier

You can find the built artifact here:

https://github.com/Thomas-Avery/test-rustls-platform-verifier/actions/runs/10378079748

The error when running this artifact:

```
PS C:\Users\Colton\Desktop> .\test-rustls-platform-verifier.exe
[2024-08-22T17:19:28Z ERROR rustls_platform_verifier::verification::windows] failed to verify TLS certificate: invalid peer certificate: InvalidPurpose
Error: reqwest::Error { kind: Request, url: Url { scheme: "https", cannot_be_a_base: false, username: "", password: None, host: Some(Domain("httpbin.org")), port: None, path: "/ip", query: None, fragment: None }, source: hyper_util::client::legacy::Error(Connect, Custom { kind: Other, error: Custom { kind: InvalidData, error: InvalidCertificate(InvalidPurpose) } }) }
```

error

complexspaces:
This PR implements two fixes in the Windows verifier to fix UB and future-UB:
- Properly `NULL`-terminate the EKU strings we pass in.
- Only pass in a thin pointer containing the string data to Windows, not the slice's fat pointer.

See the linked issue for a more detailed analysis if your curious why this was never problematic before.

Closes #126

### DIFF
--- a/rustls-platform-verifier/src/verification/mod.rs
+++ b/rustls-platform-verifier/src/verification/mod.rs
@@ -82,8 +82,11 @@ fn invalid_certificate(reason: impl Into<String>) -> rustls::Error {
 /// - id-kp-serverAuth
 // TODO: Chromium also allows for `OID_ANY_EKU` on Android.
 #[cfg(target_os = "windows")]
+// XXX: Windows requires that we NUL terminate EKU strings and we want to make sure that only the
+// data part of the `&str` pointer (using `.as_ptr()`), not all of its metadata.
+// This can be cleaned up when our MSRV is increased to 1.77 and C-string literals are available.
+// See https://github.com/rustls/rustls-platform-verifier/issues/126#issuecomment-2306232794.
 const ALLOWED_EKUS: &[*mut u8] = &["1.3.6.1.5.5.7.3.1\0".as_ptr() as *mut u8];
-
 #[cfg(target_os = "android")]
 pub const ALLOWED_EKUS: &[&str] = &["1.3.6.1.5.5.7.3.1"];
 
--- a/rustls-platform-verifier/src/verification/mod.rs
+++ b/rustls-platform-verifier/src/verification/mod.rs
@@ -74,7 +74,6 @@ fn invalid_certificate(reason: impl Into<String>) -> rustls::Error {
     )))
 }
 
-#[cfg(any(windows, target_os = "android"))]
 /// List of EKUs that one or more of that *must* be in the end-entity certificate.
 ///
 /// Legacy server-gated crypto OIDs are assumed to no longer be in use.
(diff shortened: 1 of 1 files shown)


complexspaces:
@ctz I meant to open a PR against the [rel-0.1](https://github.com/rustls/rustls-platform-verifier/commits/rel-0.1/) branch with this backport since its UB/important bugfix but I mistakenly pushed this commit to the branch instead. I apologize for that. Please make sure it matches this for diligence as part of your review.

rustls-platform-verifier/src/verification/mod.rs
```diff
@@ -74,14 +74,17 @@ fn invalid_certificate(reason: impl Into<String>) -> rustls::Error {
...
 
-#[cfg(any(windows, target_os = "android"))]
 /// List of EKUs that one or more of that *must* be in the end-entity certificate.
 ///
 /// Legacy server-gated crypto OIDs are assumed to no longer be in use.
 ///
 /// Currently supported:
 /// - id-kp-serverAuth
 // TODO: Chromium also allows for `OID_ANY_EKU` on Android.
+#[cfg(target_os = "windows")]
```
djc, line 84: If we have to declare these separately, maybe it makes more sense to move them into `android.rs` and `windows.rs`?

Also, maybe this is worth two lines of comments explaining the context here and/or linking to the issue?

djc (approved):

rustls-platform-verifier/src/verification/mod.rs
```diff
@@ -74,14 +74,17 @@ fn invalid_certificate(reason: impl Into<String>) -> rustls::Error {
...
+const ALLOWED_EKUS: &[*mut u8] = &["1.3.6.1.5.5.7.3.1\0".as_ptr() as *mut u8];
```
ctz, line 85: 💭 raised #128 containing an idea on making this clearer.

ctz: 💭 also when our msrv allows it, we could use a `c""` string here

rustls-platform-verifier/src/verification/mod.rs
complexspaces: I would prefer to keep these next to each other in the source since we still believe they should have the OIDs inside in sync for consistency between platforms.

ctz (approved):

rustls-platform-verifier/src/verification/mod.rs
complexspaces: Yeah c string literals would be really nice here 😅

rustls-platform-verifier/src/verification/mod.rs
complexspaces: I've updated this code to include a link to the issue and context around why the string handling looks the way it does.

rustls-platform-verifier/src/verification/mod.rs
```diff
@@ -82,8 +82,11 @@ fn invalid_certificate(reason: impl Into<String>) -> rustls::Error {
 /// - id-kp-serverAuth
 // TODO: Chromium also allows for `OID_ANY_EKU` on Android.
 #[cfg(target_os = "windows")]
+// XXX: Windows requires that we NUL terminate EKU strings and we want to make sure that only the
```
djc, line 85: I think this sentence is missing a word. Otherwise LGTM!

complexspaces: Oops, good catch. I'll fix this when migrating to `windows_sys`.

PR #129, opened 2024-08-23
Prepare 0.3.4

ctz:


djc (approved):
Are we going to do releases for 0.1 and 0.2, too? Should we yank older releases?

complexspaces:
I'm in favor of yanking older releases 👍

> releases for 0.1 and 0.2, too?

#130 is covering 0.1 (which is what 1P internally uses). I lean towards not patching 0.2 since the crates.io downloads of it seem basically at background noise/bot levels and everyone else is on `0.3`.

complexspaces (approved):

djc:
@complexspaces what's blocking 1P from upgrading to newer rustls?

complexspaces:
> what's blocking 1P from upgrading to newer rustls?

*ring* for better or worse. We are still stuck on an alpha `0.17` of *ring* with RSA-OEAP stuff included in it and have not spent significant time internally either migrating to `rsa` (which has vulns in it at the moment) or contributing upstream to _ring_ to get those changes into a real `0.17` release 😓.

ctz:
See [v/0.3.4](https://github.com/rustls/rustls-platform-verifier/releases/tag/v%2F0.3.4) and https://crates.io/crates/rustls-platform-verifier/0.3.4

As per the previous release PR, I checked there were no changes in android/ since the last release.

Issue #58, opened 2024-01-05
Consider adding Verifier::new_with_extra_roots implementation to other platforms


complexspaces:
The functionality of `new_with_extra_roots` is primarily useful for Linux/WASM/BSD platforms that don't have a consistent source of trusted CA root/anchors available. However, many private/internal applications often use their own private CAs instead of publicly issued ones. This seems like a use case we could support without much burden, even if those users might be better off making their own `webpki`-based verifier instead.

## Implementation details

It's worth noting that the Apple and Windows code for this already exists in a near-drop in form. Android would require more work to make a `TrustManager` that combined certificates.

### macOS/iOS

We can call [`SecTrustSetAnchorCertificates`](https://developer.apple.com/documentation/security/1396098-sectrustsetanchorcertificates?language=objc) to add additional roots to the evaluation, and then call [`SecTrustSetAnchorCertificatesOnly`](https://developer.apple.com/documentation/security/1399071-sectrustsetanchorcertificatesonl?language=objc) with `false` to trust both the custom roots and the default OS-provided ones.

### Windows

We can create a custom `CERT_CHAIN_ENGINE_CONFIG` and set `cAdditionalStore` to a custom, in-memory certificate store. This engine and store are then included in our call to `CertGetCertificateChain`.

### Android

I believe this can be done by using a [`PKIXParameters`](https://developer.android.com/reference/java/security/cert/PKIXParameters#PKIXParameters(java.util.Set%3Cjava.security.cert.TrustAnchor%3E)). I think the best idea is to create a custom `TrustManager` class that considers the system `Keystore`'s trustmanager and then a custom `Keystore` containing the user-provided roots. I'm not yet sure about the exact implementation strategy though since the Android X509 and PKIX APIs are a handful. [This](https://github.com/moxie0/AndroidPinning/blob/master/src/org/thoughtcrime/ssl/pinning/PinningTrustManager.java) and [this blog](https://codyaray.com/2013/04/java-ssl-with-multiple-keystores) may be helpful references.

blind-oracle:
I second that. This also makes development easier since when using MacOS I've spent some time figuring out why the `new_with_extra_roots` method is missing :)

djc:
Note that this is blocking (built-in) support for rustls-platform-verifier in reqwest: https://github.com/seanmonstar/reqwest/pull/2286#discussion_r1609846678.

complexspaces:
[...]

PR #133, opened 2024-08-26
Add new with extra roots on macos/ios

stormshield-gt:
Add support of extra roots for the moment on *macos* and *ios*. I plan to add *windows* and *android* but I believe this is already ready for initial feedback.
The work is mostly based on @complexspaces branch https://github.com/rustls/rustls-platform-verifier/tree/shared/extra-roots-additions.

I've changed the signature of new_with_extra_roots (breaking) to be uniform on all platforms.

Partially fix https://github.com/rustls/rustls-platform-verifier/issues/58

### DIFF
--- a/rustls-platform-verifier/Cargo.toml
+++ b/rustls-platform-verifier/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "rustls-platform-verifier"
-version = "0.3.4"
+version = "0.4.0"
 authors = ["ComplexSpaces <complexspacescode@gmail.com>", "1Password"]
 description = "rustls-platform-verifier supports verifying TLS certificates in rustls with the operating system verifier"
 keywords = ["tls", "certificate", "verification", "os", "native"]
--- a/rustls-platform-verifier/Cargo.toml
+++ b/rustls-platform-verifier/Cargo.toml
@@ -19,7 +19,7 @@ crate-type = ["cdylib", "rlib"]
 # Enables a C interface to use for testing where `cargo` can't be used.
 # This feature is not stable, nor is the interface exported when it is enabled.
 # Do not rely on this or use it in production.
-ffi-testing = ["android_logger", "rustls/ring"]
+ffi-testing = ["android_logger", "rustls/ring", "paste"]
 # Enables APIs that expose lower-level verifier types for debugging purposes.
 dbg = []
 # Enables `log::debug` base64-encoded logging of all end-entity certificates processed
@@ -34,6 +34,7 @@ log = { version = "0.4" }
 base64 = { version = "0.22", optional = true } # Only used when the `cert-logging` feature is enabled.
 jni = { version = "0.19", default-features = false, optional = true } # Only used during doc generation
 once_cell = "1.9"
+paste = { version = "1.0", default-features = false, optional = true } # Only used when `ffi-testing` feature is enabled
 
 [target.'cfg(all(unix, not(target_os = "android"), not(target_os = "macos"), not(target_os = "ios"), not(target_os = "tvos"), not(target_arch = "wasm32")))'.dependencies]
 rustls-native-certs = "0.7"
@@ -64,6 +65,7 @@ windows-sys = { version = "0.52", default-features = false, features = ["Win32_F
 
 [dev-dependencies]
 rustls = { version = "0.23", default-features = false, features = ["ring"] }
+paste = { version = "1.0", default-features = false } # Only used when `ffi-testing` feature is enabled
 
 [package.metadata.docs.rs]
 rustdoc-args = ["--cfg", "docsrs"]
--- a/rustls-platform-verifier/Cargo.toml
+++ b/rustls-platform-verifier/Cargo.toml
@@ -46,12 +46,12 @@ webpki = { package = "rustls-webpki", version = "0.102", default-features = fals
 android_logger = { version = "0.13", optional = true } # Only used during testing.
 
 [target.'cfg(target_arch = "wasm32")'.dependencies]
-webpki-roots = "0.26"
 webpki = { package = "rustls-webpki", version = "0.102", default-features = false }
+webpki-root-certs = "0.26"
 
-# BSD targets require webpki-roots for the real-world verification tests.
+# BSD targets require webpki-roots-certs for the real-world verification tests.
 [target.'cfg(target_os = "freebsd")'.dev-dependencies]
-webpki-roots = "0.26"
+webpki-root-certs = "0.26"
 
 [target.'cfg(any(target_os = "macos", target_os = "ios", target_os = "tvos"))'.dependencies]
 core-foundation = "0.9"
--- a/rustls-platform-verifier/src/tests/verification_mock/mod.rs
+++ b/rustls-platform-verifier/src/tests/verification_mock/mod.rs
@@ -221,7 +221,7 @@ mock_root_test_cases! {
     // with AIA because there's no AIA issuer field in the certificate).
     // (AIA is an extension that allows downloading of missing data,
     // like missing certificates, during validation; see
-    // https://datatracker.ietf.org/doc/html/rfc5280#section-5.2.7).
+    // https://datatracker.ietf.org/doc/html/rfc5280#section-4.2.2.1).
     ee_only_dns [ any(windows, unix) ] => TestCase {
         reference_id: EXAMPLE_COM,
         chain: &[ROOT1_INT1_EXAMPLE_COM_GOOD],
--- a/rustls-platform-verifier/src/tests/verification_mock/mod.rs
+++ b/rustls-platform-verifier/src/tests/verification_mock/mod.rs
@@ -41,6 +41,14 @@ macro_rules! mock_root_test_cases {
                 pub fn $name() {
                     super::$name()
                 }
+
+                paste::paste!{
+                    #[cfg(all($target, not(windows), not(target_os = "android")))]
+                    #[test]
+                    pub fn [<$name _extra>](){
+                        super::[<$name _extra>]()
+                    }
+                }
             )+
 
         }
@@ -49,8 +57,15 @@ macro_rules! mock_root_test_cases {
         pub static ALL_TEST_CASES: &'static [fn()] = &[
             $(
                 #[cfg($target)]
-                $name
+                $name,
+
+                paste::paste!{
+                    #[cfg(all($target, not(windows), not(target_os = "android")))]
+                    [<$name _extra>]
+                }
+
             ),+
+
         ];
     };
 
@@ -58,7 +73,14 @@ macro_rules! mock_root_test_cases {
         $(
             #[cfg($target)]
             pub(super) fn $name() {
-                test_with_mock_root(&$test_case);
+                test_with_mock_root(&$test_case, Roots::OnlyExtra);
+            }
+
+            paste::paste!{
+                #[cfg(all($target, not(windows), not(target_os = "android")))]
+                pub(super) fn [<$name _extra>]() {
+                    test_with_mock_root(&$test_case, Roots::ExtraAndPlatform);
+                }
             }
         )+
     };
@@ -301,11 +323,18 @@ mock_root_test_cases! {
     },
 }
 
-fn test_with_mock_root<E: std::error::Error + PartialEq + 'static>(test_case: &TestCase<E>) {
+fn test_with_mock_root<E: std::error::Error + PartialEq + 'static>(
+    test_case: &TestCase<E>,
+    root_src: Roots,
+) {
     ensure_global_state();
     log::info!("verifying {:?}", test_case.expected_result);
 
-    let verifier = Verifier::new_with_fake_root(ROOT1); // TODO: time
+    let verifier = match root_src {
+        Roots::OnlyExtra => Verifier::new_with_fake_root(ROOT1), // TODO: time
+        #[cfg(all(unix, not(target_os = "android")))]
+        Roots::ExtraAndPlatform => Verifier::new_with_extra_roots(vec![ROOT1.into()]),
+    };
     let mut chain = test_case
         .chain
         .iter()
@@ -336,4 +365,16 @@ fn test_with_mock_root<E: std::error::Error + PartialEq + 'static>(test_case: &T
         test_case.other_error.as_ref(),
     );
     // TODO: get into specifics of errors returned when it fails.
+}
+
+enum Roots {
+    /// Test with only extra roots, without loading the platform trust store.
+    ///
+    /// We want to keep things reproducible given the background-managed nature of trust roots on platforms.
+    OnlyExtra,
+    /// Test with loading the extra roots and the platform trust store.
+    ///
+    /// Right now, not all platforms are supported.
+    #[cfg(all(unix, not(target_os = "android")))]
+    ExtraAndPlatform,
 }
--- a/rustls-platform-verifier/src/tests/verification_mock/mod.rs
+++ b/rustls-platform-verifier/src/tests/verification_mock/mod.rs
@@ -86,10 +86,11 @@ pub(super) fn verification_without_mock_root() {
     ensure_global_state();
     // Since Rustls 0.22 constructing a webpki verifier (like the one backing Verifier on unix
     // systems) without any roots produces `OtherError(NoRootAnchors)` - since our FreeBSD CI
-    // runner fails to find any roots with openssl-probe we need to provide webpki-roots here
+    // runner fails to find any roots with openssl-probe we need to provide webpki-root-certs here
     // or the test will fail with the `OtherError` instead of the expected `CertificateError`.
     #[cfg(target_os = "freebsd")]
-    let verifier = Verifier::new_with_extra_roots(webpki_roots::TLS_SERVER_ROOTS.iter().cloned());
+    let verifier =
+        Verifier::new_with_extra_roots(webpki_root_certs::TLS_SERVER_ROOT_CERTS.iter().cloned());
 
     #[cfg(not(target_os = "freebsd"))]
     let verifier = Verifier::new();
--- a/rustls-platform-verifier/src/verification/apple.rs
+++ b/rustls-platform-verifier/src/verification/apple.rs
@@ -43,6 +43,9 @@ fn system_time_to_cfdate(time: pki_types::UnixTime) -> Result<CFDate, TlsError> 
 /// A TLS certificate verifier that utilizes the Apple platform certificate facilities.
 #[derive(Debug)]
 pub struct Verifier {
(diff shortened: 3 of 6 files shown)


cpu (commented):
Thanks for picking this up! :two_hearts:  Here's some initial feedback.

rustls-platform-verifier/Cargo.toml
```diff
@@ -34,6 +34,7 @@ log = { version = "0.4" }
 base64 = { version = "0.22", optional = true } # Only used when the `cert-logging` feature is enabled.
 jni = { version = "0.19", default-features = false, optional = true } # Only used during doc generation
 once_cell = "1.9"
+paste = "1.0.15"
```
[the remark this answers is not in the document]
complexspaces: We might be able to gate this dependency behind either `[dev-dependencies] & #[cfg(test)]` and/or `feature = ["ffi-testing"]` to at least prevent it from being built unnecessarily downstream. Its only used in integration tests.

rustls-platform-verifier/src/tests/verification_mock/mod.rs
```diff
@@ -336,3 +366,9 @@ fn test_with_mock_root<E: std::error::Error + PartialEq + 'static>(test_case: &T
     );
     // TODO: get into specifics of errors returned when it fails.
 }
+
+enum Roots {
```
complexspaces, line 370: Do you mind documenting the rationale behind this enum? At a glance its not really clear why it needs to be platform specific or the tradeoffs between it and changing `test_only_root_ca_override` to just be a boolean that these tests set instead.

rustls-platform-verifier/src/verification/others.rs
```diff
@@ -54,11 +54,17 @@ impl Verifier {
...
-        roots: impl IntoIterator<Item = pki_types::TrustAnchor<'static>>,
+        roots: impl IntoIterator<Item = impl Into<pki_types::CertificateDer<'static>>>,
     ) -> Self {
         Self {
             inner: OnceCell::new(),
-            extra_roots: roots.into_iter().collect::<Vec<_>>().into(),
+            extra_roots: roots
+                .into_iter()
+                .flat_map(|root| {
+                    webpki::anchor_from_trusted_cert(&root.into()).map(|anchor| anchor.to_owned())
```
complexspaces, line 64: Is this `to_owned()` call needed? AFAICT `anchor_from_trusted_cert` should be remembering the `'static` lifetime associated with the `CertificateDer<'static>` and returning an "owned" one already.

complexspaces:
As a Git nit, would you mind adding me as a `Co-authored-by:` to 1bb8c1c0551bea10af23400029e9d0f4dfd14178 since its based off my original branch?

>  I plan to add windows and android 

I planned on working on overhauling the Android verifier this week since it has a lot of outstanding problems as-is, including not supporting this pattern very cleanly. Maybe we could split the Windows and Android work?

cpu:
> Maybe we could split the Windows and Android work?

That sounds like a good idea to me :+1:

stormshield-gt:
> As a Git nit, would you mind adding me as a `Co-authored-by:` to [1bb8c1c](https://github.com/rustls/rustls-platform-verifier/commit/1bb8c1c0551bea10af23400029e9d0f4dfd14178) since its based off my original branch?

Sure ! For what it's worth, I've already done it but it seems the change has been reverted, most likely because of a rebase

> I planned on working on overhauling the Android verifier this week since it has a lot of outstanding problems as-is, including not supporting this pattern very cleanly. Maybe we could split the Windows and Android work?

All right, I will look into the windows implementation then. Maybe I will just do a rebase over https://github.com/rustls/rustls-platform-verifier/pull/131 to start cleanly.

rustls-platform-verifier/src/verification/others.rs
stormshield-gt: If I try to remove the `to_owned` I get a 
```sh
error[E0515]: cannot return value referencing temporary value
  --> rustls-platform-verifier/src/verification/others.rs:63:34
   |
63 |                 .flat_map(|root| webpki::anchor_from_trusted_cert(&root.into()))
   |                                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^-----------^
   |                                  |                                 |
   |                                  |                                 temporary value created here
   |                                  returns a value referencing data owned by the current function 
```

I guess it's because `webpki::anchor_from_trusted_cert` doesn't take ownership of the certificate, so the even if `CertificateDer` have a `'static` it would be desallocated at the end of the closure so we can't take a reference to it

rustls-platform-verifier/src/tests/verification_mock/mod.rs
stormshield-gt: Thanks for the feedback, I've added some comments on the variant, I hope that better justify the `enum`

> changing test_only_root_ca_override to just be a boolean that these tests set instead.

It could totally be a boolean, as personal preference, I tend to avoid boolean parameters because it's not readable at call site. Especially it environment like GitHub where you don't have inlay hint, a function with boolean parameters can soon become cryptic, for instance `foo(true, false)/foo(false, true)`.

Anyway, as there is only one boolean parameter this not as impactful, I can change it to a `bool` if you want.

> its not really clear why it needs to be platform specific

I think we will get rid of this limitation when we support `new_with_extra_roots`  on all the tested platforms.

rustls-platform-verifier/Cargo.toml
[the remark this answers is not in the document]
stormshield-gt: Thanks for the suggestion, indeed I've been able to gate this feature behind `[dev-dependencies] & #[cfg(test)]`

rustls-platform-verifier/src/tests/verification_mock/mod.rs
```diff
@@ -221,7 +221,7 @@ mock_root_test_cases! {
     // with AIA because there's no AIA issuer field in the certificate).
     // (AIA is an extension that allows downloading of missing data,
     // like missing certificates, during validation; see
-    // https://datatracker.ietf.org/doc/html/rfc5280#section-5.2.7).
```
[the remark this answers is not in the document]
stormshield-gt: Yes sure, I've split this in a dedicated commit.

rustls-platform-verifier/src/verification/apple.rs
```diff
@@ -43,21 +43,43 @@ fn system_time_to_cfdate(time: pki_types::UnixTime) -> Result<CFDate, TlsError>
...
+    /// facilities with the addition of extra root certificates to trust.
+    ///
+    /// See [Verifier::new] for the external requirements the verifier needs.
+    pub fn new_with_extra_roots(
+        roots: impl IntoIterator<Item = impl Into<pki_types::CertificateDer<'static>>>,
+    ) -> Self {
+        Self {
+            extra_roots: roots
+                .into_iter()
+                .map(|root| root.into())
```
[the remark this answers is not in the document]
stormshield-gt: Thanks, much cleaner that way

rustls-platform-verifier/src/verification/apple.rs
```diff
@@ -123,21 +146,49 @@ impl Verifier {
...
         }
 
-        // When testing, support using fake roots and ignoring values present on the system.
+        let raw_extra_roots = self.extra_roots.iter();
+
+        #[cfg(any(test, feature = "ffi-testing", feature = "dbg"))]
+        let extra_root = self
+            .test_only_root_ca_override
+            .as_ref()
+            .map(|root| pki_types::CertificateDer::from(root.as_slice().to_owned()));
```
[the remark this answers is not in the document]
stormshield-gt: Thanks, that a lot more concise

cpu (approved):
Thanks! This looks good to me. Did you want to keep it as a draft until https://github.com/rustls/rustls-platform-verifier/pull/131 is merged?

stormshield-gt:
I've rebased on top of https://github.com/rustls/rustls-platform-verifier/pull/131 and tried to implement the `new_with_extra_roots` for `Windows`, thanks to the guidance of @complexspaces 

> We can create a custom CERT_CHAIN_ENGINE_CONFIG and set cAdditionalStore to a custom, in-memory certificate store. This engine and store are then included in our call to CertGetCertificateChain.

Sadly, it keeps considering the custom CA as untrusted. I've tried to tweak the flags and various parameters, and add debug traces but without success. Anyone has any idea or how we can solve this ?

On GitHub there are very few usages of `rghAdditionalStore/cAdditionalStore`, most of them seems to use the same parameters as me. For instance [here](https://github.com/wqweto/VbAsyncSocket/blob/f91a05203831cc070ef864ca02844178f7cb38d4/src/cTlsSocket.cls#L2718
) and [here](https://github.com/akamai/akamai-security-research/blob/52e4920ddd4f5c8eb5f498a73085c8980a208d1b/PoCs/CVE-2022-34689/vulnerability_local_demo.cpp#L61)

cpu:
> I've rebased on top of https://github.com/rustls/rustls-platform-verifier/pull/131 and tried to implement the new_with_extra_roots for Windows

I think it would be better to land the MacOS code and handle Windows as a follow-up, especially if there are implementation issues to sort out. The MacOS code seemed ready mod review comments. WDYT?

stormshield-gt:
Yes sure, I will split the Windows work on an other PR

rustls-platform-verifier/src/verification/apple.rs
```diff
@@ -43,21 +43,40 @@ fn system_time_to_cfdate(time: pki_types::UnixTime) -> Result<CFDate, TlsError>
...
+            crypto_provider: OnceCell::new(),
+        }
+    }
+
+    /// Creates a new instance of a TLS certificate verifier that utilizes the Apple certificate
+    /// facilities with the addition of extra root certificates to trust.
+    ///
+    /// See [Verifier::new] for the external requirements the verifier needs.
+    pub fn new_with_extra_roots(
+        roots: impl IntoIterator<Item = impl Into<pki_types::CertificateDer<'static>>>,
```
djc, line 76: I would prefer to avoid the `Into` indirection here. `CertificateDer` implements `From` for `&[u8]` but IMO the type safety goals of pki-types are kind of defeated if we don't require the caller to handle that conversion explicitly. (As far as I can tell we don't do this in other places.)

rustls-platform-verifier/src/verification/apple.rs
```diff
@@ -123,21 +143,49 @@ impl Verifier {
...
+
+        // If any extra roots were provided by the user (or tests), provide them to the trust
+        // evaluation regardless of their system trust settings or status.
+        if !extra_roots.is_empty() {
+            trust_evaluation
+                .set_anchor_certificates(&extra_roots)
+                .map_err(|e| TlsError::Other(OtherError(Arc::new(e))))?;
+
+            // We want to trust both the system-installed and the extra roots. This must be set
+            // since by calling `SecTrustSetAnchorCertificates` "disables the trusting of any
```
djc, line 172: This comment reads a little awkwardly? Suggest removing "by".

rustls-platform-verifier/src/verification/others.rs
```diff
@@ -54,11 +54,17 @@ impl Verifier {
     /// WebPKI, using root certificates provided by the platform and augmented by
     /// the provided extra root certificates.
     pub fn new_with_extra_roots(
-        roots: impl IntoIterator<Item = pki_types::TrustAnchor<'static>>,
+        roots: impl IntoIterator<Item = impl Into<pki_types::CertificateDer<'static>>>,
```
djc, line 57: As for macOS, I prefer skipping the `Into` indrection here.

Also noteworthy that this breaks compatibility. @complexspaces do you plan on backporting this stuff to 0.1?

djc (approved):

rustls-platform-verifier/src/verification/apple.rs
stormshield-gt: That's clearer indeed, I removed the `by`

rustls-platform-verifier/src/verification/apple.rs
stormshield-gt: Thanks for the feedback, I see your point, I've removed the extra `Into`

rustls-platform-verifier/src/verification/others.rs
stormshield-gt: > (Maybe tack on a commit that bumps the version number to 0.4?)

Thanks for the suggestion, that's done

cpu: We should consider adding cargo-semver-check to this repo's CI. It looks like we haven't done that yet. It's not perfect, but pretty good at flagging this sort of change.

rustls-platform-verifier/src/verification/others.rs
```diff
@@ -154,7 +160,7 @@ impl Verifier {
 
         #[cfg(target_arch = "wasm32")]
         {
-            root_store.add_trust_anchors(webpki_roots::TLS_SERVER_ROOTS.iter().map(|root| {
+            root_store.add_trust_anchors(webpki_roots::TLS_SERVER_ROOT_CERTS.iter().map(|root| {
```
djc, line 163: This looks wrong? If this needs to be `TLS_SERVER_ROOT_CERTS`, it should come from the `webpki_root_certs` crate, right? Are we not compiling for `wasm32` in CI?

stormshield-gt: [wasm isn't checked on the CI](https://github.com/rustls/rustls-platform-verifier/blob/8080972e5fb8f5b5aeb0172abac18d5834d11ced/.github/workflows/ci.yml#L70-L75) and seems broken on main, at least on my computer. The function `add_trust_anchors` is not even in `rustls` anymore. 
Maybe we should fill a separate issue for wasm? I've changed to `webpki_root_certs` as you rightfully pointed out, but this won't be sufficient to fix the whole target.

rustls-platform-verifier/src/verification/others.rs
complexspaces: +1 for adding `cargo-semver-check` here. It runs pretty fast in CI so it shouldn't have any downsides even if it doesn't catch everything.

> . @complexspaces do you plan on backporting this stuff to 0.1?

I'm not sure if that's possible under the same SemVer release series. AFAIK 1P has no need for extra roots on the other platforms so I don't think its worth trying to figure out a workaround for either. Does that make sense?

complexspaces (approved):
Thank you for getting started on this implementation.

rustls-platform-verifier/src/verification/apple.rs
```diff
@@ -43,21 +43,40 @@ fn system_time_to_cfdate(time: pki_types::UnixTime) -> Result<CFDate, TlsError>
 /// A TLS certificate verifier that utilizes the Apple platform certificate facilities.
 #[derive(Debug)]
 pub struct Verifier {
+    // Extra trust anchors to add to the verifier above and beyond those provided by the
+    // platform via rustls-native-certs.
```
complexspaces, line 47: Sorry, one nit I missed: We don't use `rustls-native-certs` on Apple platforms. This should say "system-provided trust stores" or similar instead.

rustls-platform-verifier/src/verification/others.rs
djc: Yeah, was wondering if there was a need at 1P.

rustls-platform-verifier/src/verification/apple.rs
stormshield-gt: Oh yeah I missed that too, I reword the comment

rustls-platform-verifier/src/verification/others.rs
```diff
@@ -54,11 +54,17 @@ impl Verifier {
     /// WebPKI, using root certificates provided by the platform and augmented by
     /// the provided extra root certificates.
     pub fn new_with_extra_roots(
-        roots: impl IntoIterator<Item = pki_types::TrustAnchor<'static>>,
+        roots: impl IntoIterator<Item = pki_types::CertificateDer<'static>>,
```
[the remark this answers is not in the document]
djc: I think the tricky part here would be that the conversion from `CertificateDer` to `TrustAnchor` is a non-trivial amount of DER parsing that currently lives in webpki. So far, pki-types contains very little logic, and my intuition is that it might be better to keep it that way?

rustls-platform-verifier/src/verification/apple.rs
```diff
@@ -43,21 +43,40 @@ fn system_time_to_cfdate(time: pki_types::UnixTime) -> Result<CFDate, TlsError>
...
+            crypto_provider: OnceCell::new(),
+        }
+    }
+
+    /// Creates a new instance of a TLS certificate verifier that utilizes the Apple certificate
+    /// facilities with the addition of extra root certificates to trust.
+    ///
+    /// See [Verifier::new] for the external requirements the verifier needs.
+    pub fn new_with_extra_roots(
+        roots: impl IntoIterator<Item = pki_types::CertificateDer<'static>>,
```
djc, line 76: Sorry, one more nit: I think we should just take `Vec` here as the argument? We're going to store a `Vec` internally anyway, and IMO it's slightly better to allow the caller to provide that for the case where they already have a `Vec` allocated for some reason.

(Should then do the same thing for the Unix platform in the later commit.)

djc (approved):

rustls-platform-verifier/src/verification/apple.rs
stormshield-gt: I see your point, I've updated the code

djc: Hmm, I failed to consider that on Unix we have to iterate anyway to map from `CertificateDer` to `TrustAnchor`, so I guess the optimal scenario for macOS/iOS is different from the Unices? Not sure what the best choice here is. Going to merge this now -- thanks for all the work, I am happy to do a follow-up here to change it back it to iterators if people think that would be better.

rustls-platform-verifier/src/verification/others.rs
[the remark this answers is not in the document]
complexspaces: That seems like a reasonable middleground. With that this could be a bit cleaner and push the bulkiness of roots further away from this crate (and probably others)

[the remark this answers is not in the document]
djc: @ctz @cpu thoughts?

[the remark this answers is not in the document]
ctz: I think I would prefer to allow composition of verifiers at the rustls level, rather than relax the type safety of `TrustAnchor`.

But, then, I think that's because I'm not 100% sold on the value of giving extra roots to the platform verifier (I previously thought it was a method for affixing the behaviour for testing purposes, but I guess not?)

ctz (commented):
oops, i didn't submit this

PR #135, opened 2024-08-28
Add new with extra roots on windows

stormshield-gt:
Windows implementation of `new_with_extra_roots`, which have some problems for now as mentioned in https://github.com/rustls/rustls-platform-verifier/pull/133#issuecomment-2315871410.
Note that it's based on top of https://github.com/rustls/rustls-platform-verifier/pull/133 to benefit the tests defined here.

Partially Fix https://github.com/rustls/rustls-platform-verifier/issues/58

### DIFF
--- a/rustls-platform-verifier/src/tests/verification_mock/mod.rs
+++ b/rustls-platform-verifier/src/tests/verification_mock/mod.rs
@@ -116,7 +116,8 @@ pub(super) fn verification_without_mock_root() {
             .iter()
             .cloned()
             .collect(),
-    );
+    )
+    .unwrap();
 
     #[cfg(not(target_os = "freebsd"))]
     let verifier = Verifier::new();
@@ -337,7 +338,7 @@ fn test_with_mock_root<E: std::error::Error + PartialEq + 'static>(
     let verifier = match root_src {
         Roots::OnlyExtra => Verifier::new_with_fake_root(ROOT1), // TODO: time
         #[cfg(not(target_os = "android"))]
-        Roots::ExtraAndPlatform => Verifier::new_with_extra_roots(vec![ROOT1.into()]),
+        Roots::ExtraAndPlatform => Verifier::new_with_extra_roots(vec![ROOT1.into()]).unwrap(),
     };
     let mut chain = test_case
         .chain
--- a/rustls-platform-verifier/src/tests/verification_mock/mod.rs
+++ b/rustls-platform-verifier/src/tests/verification_mock/mod.rs
@@ -43,7 +43,7 @@ macro_rules! mock_root_test_cases {
                 }
 
                 paste::paste!{
-                    #[cfg(all($target, not(windows), not(target_os = "android")))]
+                    #[cfg(all($target, not(target_os = "android")))]
                     #[test]
                     pub fn [<$name _extra>](){
                         super::[<$name _extra>]()
@@ -60,7 +60,7 @@ macro_rules! mock_root_test_cases {
                 $name,
 
                 paste::paste!{
-                    #[cfg(all($target, not(windows), not(target_os = "android")))]
+                    #[cfg(all($target, not(target_os = "android")))]
                     [<$name _extra>]
                 }
 
@@ -77,7 +77,7 @@ macro_rules! mock_root_test_cases {
             }
 
             paste::paste!{
-                #[cfg(all($target, not(windows), not(target_os = "android")))]
+                #[cfg(all($target, not(target_os = "android")))]
                 pub(super) fn [<$name _extra>]() {
                     test_with_mock_root(&$test_case, Roots::ExtraAndPlatform);
                 }
@@ -336,7 +336,7 @@ fn test_with_mock_root<E: std::error::Error + PartialEq + 'static>(
 
     let verifier = match root_src {
         Roots::OnlyExtra => Verifier::new_with_fake_root(ROOT1), // TODO: time
-        #[cfg(all(unix, not(target_os = "android")))]
+        #[cfg(not(target_os = "android"))]
         Roots::ExtraAndPlatform => Verifier::new_with_extra_roots(vec![ROOT1.into()]),
     };
     let mut chain = test_case
@@ -379,6 +379,6 @@ enum Roots {
     /// Test with loading the extra roots and the platform trust store.
     ///
     /// Right now, not all platforms are supported.
-    #[cfg(all(unix, not(target_os = "android")))]
+    #[cfg(not(target_os = "android"))]
     ExtraAndPlatform,
 }
--- a/rustls-platform-verifier/src/verification/apple.rs
+++ b/rustls-platform-verifier/src/verification/apple.rs
@@ -45,10 +45,10 @@ fn system_time_to_cfdate(time: pki_types::UnixTime) -> Result<CFDate, TlsError> 
 pub struct Verifier {
     /// Extra trust anchors to add to the verifier above and beyond those provided by
     /// the system-provided trust stores.
-    extra_roots: Vec<pki_types::CertificateDer<'static>>,
+    extra_roots: Vec<SecCertificate>,
     /// Testing only: The root CA certificate to trust.
     #[cfg(any(test, feature = "ffi-testing", feature = "dbg"))]
-    test_only_root_ca_override: Option<Vec<u8>>,
+    test_only_root_ca_override: Option<SecCertificate>,
     pub(super) crypto_provider: OnceCell<Arc<CryptoProvider>>,
 }
 
@@ -72,13 +72,22 @@ impl Verifier {
     /// facilities with the addition of extra root certificates to trust.
     ///
     /// See [Verifier::new] for the external requirements the verifier needs.
-    pub fn new_with_extra_roots(roots: Vec<pki_types::CertificateDer<'static>>) -> Self {
-        Self {
-            extra_roots: roots,
+    pub fn new_with_extra_roots(
+        roots: Vec<pki_types::CertificateDer<'static>>,
+    ) -> Result<Self, TlsError> {
+        let extra_roots = roots
+            .into_iter()
+            .map(|root| {
+                SecCertificate::from_der(&root)
+                    .map_err(|_| TlsError::InvalidCertificate(CertificateError::BadEncoding))
+            })
+            .collect::<Result<Vec<_>, _>>()?;
+        Ok(Self {
+            extra_roots,
             #[cfg(any(test, feature = "ffi-testing", feature = "dbg"))]
             test_only_root_ca_override: None,
             crypto_provider: OnceCell::new(),
-        }
+        })
     }
 
     /// Creates a test-only TLS certificate verifier which trusts our fake root CA cert.
@@ -86,7 +95,7 @@ impl Verifier {
     pub(crate) fn new_with_fake_root(root: &[u8]) -> Self {
         Self {
             extra_roots: Vec::new(),
-            test_only_root_ca_override: Some(root.into()),
+            test_only_root_ca_override: Some(SecCertificate::from_der(root).unwrap()),
             crypto_provider: OnceCell::new(),
         }
     }
@@ -141,29 +150,24 @@ impl Verifier {
                 .map_err(|e| invalid_certificate(e.to_string()))?;
         }
 
-        let raw_extra_roots = self.extra_roots.iter();
+        #[cfg(not(any(test, feature = "ffi-testing", feature = "dbg")))]
+        let extra_roots = self.extra_roots.as_slice();
 
         #[cfg(any(test, feature = "ffi-testing", feature = "dbg"))]
-        let extra_root = self
-            .test_only_root_ca_override
-            .as_ref()
-            .map(|root| pki_types::CertificateDer::from_slice(root));
-
+        let extra_roots: Vec<_> = self
+            .extra_roots
+            .iter()
+            .chain(self.test_only_root_ca_override.as_ref())
+            .cloned()
+            .collect();
         #[cfg(any(test, feature = "ffi-testing", feature = "dbg"))]
-        let raw_extra_roots = raw_extra_roots.chain(&extra_root).to_owned();
-
-        let extra_roots = raw_extra_roots
-            .map(|root| {
-                SecCertificate::from_der(root)
-                    .map_err(|_| TlsError::InvalidCertificate(CertificateError::BadEncoding))
-            })
-            .collect::<Result<Vec<_>, _>>()?;
+        let extra_roots = extra_roots.as_slice();
 
         // If any extra roots were provided by the user (or tests), provide them to the trust
         // evaluation regardless of their system trust settings or status.
         if !extra_roots.is_empty() {
             trust_evaluation
-                .set_anchor_certificates(&extra_roots)
+                .set_anchor_certificates(extra_roots)
                 .map_err(|e| TlsError::Other(OtherError(Arc::new(e))))?;
 
             // We want to trust both the system-installed and the extra roots. This must be set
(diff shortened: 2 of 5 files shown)


stormshield-gt:
After experimenting using [`cAdditionalStore/rghAdditionalStore`](https://learn.microsoft.com/en-us/windows/win32/api/wincrypt/ns-wincrypt-cert_chain_engine_config#members), I concluded that this store can only be used for intermediate certificates and not for roots.
Indeed, when configuring the `certEngine`, if I add the intermediates only to the `additonalStore` and the extra roots to the `exclusiveStore`, the validation is successful. However, if I add both the intermediates and the extra roots to the `additionalStore`, then the validation failed, saying it doesn't trust the root.

So we can't use `cAdditionalStore` to achieve what we want. [Chromium seems to have also dropped this idea](https://issues.chromium.org/issues/40578169).

I've come up with 3 possible solutions: 

1) Use `hExclusiveRoot` to load the system trust store and the extra roots. I got a POC locally that pass the tests that looks like this:
<details>

<summary>Click to expand the code</summary> 


```rust
    fn new_with_extra_roots(
        roots: &[pki_types::CertificateDer<'static>],
    ) -> Result<Self, TlsError> {

        let mut inner = Self::new()?;

        // In memory store for the extra roots
        let mut additional_store = CertificateStore::new()?;
        for root in roots {
            additional_store.add_cert(root)?;
        }

        // System store
        let mut pvpara: Vec<_> = "root".encode_utf16().collect();
        pvpara.push(0);
        let system_store = unsafe { CertOpenSystemStoreW(0, pvpara.as_ptr()) };

        // We create a collection, and we add the two previous created stores.
        let collection = unsafe {
            CertOpenStore(
                CERT_STORE_PROV_COLLECTION,
                X509_ASN_ENCODING | PKCS_7_ASN_ENCODING,
                0,
                CERT_STORE_DEFER_CLOSE_UNTIL_LAST_FREE_FLAG,
                ptr::null(), //must be null in that case
            )
        };
        unsafe {
            CertAddStoreToCollection(
                collection,
                system_store,
                CERT_PHYSICAL_STORE_ADD_ENABLE_FLAG,
                0,
            )
        };
        unsafe {
            CertAddStoreToCollection(
                collection,
                additional_store.inner.as_ptr(),
                CERT_PHYSICAL_STORE_ADD_ENABLE_FLAG,
                0,
            )
        };

        let mut config = CERT_CHAIN_ENGINE_CONFIG::zeroed_with_size();
        config.hExclusiveRoot = collection;
        //..
```

</details>

However, this seems fragile as in Windows [they are Root stores in various places](https://learn.microsoft.com/en-us/windows/win32/seccrypto/system-store-locations). It will be hard to cover and maintain the roots that match the ones that are naturally looked by [`CertGetCertificateChain`](https://learn.microsoft.com/en-us/windows/win32/api/wincrypt/nf-wincrypt-certgetcertificatechain) that don't use `exclusiveStore`.

2) Do the verification twice in case of extra roots and combine the result. One verification with only the system root store and one with only the extra roots. That's what's the Golang std seems to do, but for backward compatibility reasons. To cite the [go issue](https://github.com/golang/go/issues/46287#issue-896919270) :
> In order not to break existing code that invokes x509.SystemCertPool and then adds extra custom certificates to it, we’ll have Verify perform two verification in that case, one with the platform and one with the custom roots, and return chains from both.

3) Use the [same approach](https://github.com/steffengy/schannel-rs/blob/master/src/tls_stream.rs#L651-L774) as `schannel` crate. Basically, they add the [`CERT_CHAIN_POLICY_ALLOW_UNKNOWN_CA_FLAG`](https://learn.microsoft.com/en-us/windows/win32/api/wincrypt/ns-wincrypt-cert_chain_policy_para#members) flag if the root of the chain is in the extra roots.

### Other references

In the source code of [curl](https://github.com/curl/curl/blob/4ff04615a095ed1e7c489e3bbd4600b52e68d9a7/lib/vtls/schannel_verify.c#L598-L791) and [Dotnet](https://github.com/dotnet/runtime/blob/8d24ba37861831302a2ddf994d64626ce7e2b58a/src/libraries/System.Security.Cryptography/src/System/Security/Cryptography/X509Certificates/ChainPal.Windows.BuildChain.cs#L95), we can see that they only support using extra roots exclusively, not in combination with the system root store like we plan to support.

stormshield-gt:
This MR is ready to review, I've opted for the [golang approach](https://github.com/golang/go/blob/2707d42966f8985a6663c93e943b9a44b9399fca/src/crypto/x509/verify.go#L774-L783) that I discussed on discord.

The Android job failure seems unrelated, as it's already on `main`.

rustls-platform-verifier/src/verification/windows.rs
```diff
@@ -34,16 +34,18 @@ use std::{
     ptr::{self, NonNull},
     sync::Arc,
 };
+use windows_sys::Win32::Security::Cryptography::CERT_TRUST_IS_PARTIAL_CHAIN;
 use windows_sys::Win32::{
```
complexspaces, line 38: Nit: Merge this import into the other block for consistency?

rustls-platform-verifier/src/verification/windows.rs
```diff
@@ -456,6 +483,22 @@ impl Verifier {
...
+    ///
+    /// A [`CryptoProvider`] must be set with
+    /// [`set_provider`][Verifier::set_provider]/[`with_provider`][Verifier::with_provider] or
+    /// [`CryptoProvider::install_default`] before the verifier can be used.
+    pub fn new_with_extra_roots(roots: Vec<pki_types::CertificateDer<'static>>) -> Self {
+        Self {
+            #[cfg(any(test, feature = "ffi-testing", feature = "dbg"))]
+            test_only_root_ca_override: None,
+            crypto_provider: OnceCell::new(),
+            extra_roots: roots.into_iter().map(Into::into).collect::<Vec<_>>(),
```
complexspaces, line 501: Is this iteration/collection needed? It looks like we are just converting from `Vec<T>` to `Vec<T>` again.

rustls-platform-verifier/src/verification/windows.rs
```diff
@@ -521,7 +565,24 @@ impl Verifier {
...
-        let cert_chain = store.new_chain_in(&primary_cert, now)?;
+        let mut cert_chain = store.new_chain_in(&primary_cert, now)?;
+
+        // If we have extra roots and the building the chain gave us an error, we try to build a
+        // new one with the extra roots.
+        if !self.extra_roots.is_empty()
+            // SAFETY: The pointer is guaranteed to be non-null.
+            && unsafe { *cert_chain.inner.as_ptr() }
+                .TrustStatus
+                .dwErrorStatus & CERT_TRUST_IS_PARTIAL_CHAIN != 0
```
complexspaces, line 576: Nit: Please extract this out into its own binding for clarity. For example:
```rust
let build_status = unsafe { cert_chain.inner.as_ref().TrustStatus.dwErrorStatus };
if build_status & CERT_TRUST_IS_PARTIAL_CHAIN != 0 && !self.extra_roots.is_empty() {
    ...
}
```

rustls-platform-verifier/src/verification/windows.rs
```diff
@@ -521,7 +565,24 @@ impl Verifier {
...
 
-        let cert_chain = store.new_chain_in(&primary_cert, now)?;
+        let mut cert_chain = store.new_chain_in(&primary_cert, now)?;
+
+        // If we have extra roots and the building the chain gave us an error, we try to build a
+        // new one with the extra roots.
+        if !self.extra_roots.is_empty()
+            // SAFETY: The pointer is guaranteed to be non-null.
+            && unsafe { *cert_chain.inner.as_ptr() }
+                .TrustStatus
```
complexspaces, line 575: Please leave a comment around here to say we only use `TrustStatus` here because it hasn't had verification performed on it. I think its worthwhile to callout why we use `TrustStatus` up here but then use `dwError` instead after the chain verification.

complexspaces (changes requested):
Thanks for the PR, I think this approach is the most reasonable we have, and the code looks good for the most part.

One open question I have is about the infallibility of `new_with_extra_roots`. After seeing how we need to implement this on Windows, I'm wondering if we want to return `Result<Self, TlsError>` instead so that we can do OS-level logic in the constructor instead of being restricted to infallible operations.

The reason I mention this is because I think the re-build logic in this is not great for efficiency if we assume that the case where extra roots are provided is required for most, if not all, chain building operations to succeed. Instead we are allocating a new `CertificateStore`, parsing all the DER provided by the user again, etc. Instead, you could envision that we keep an `Option<CertificateStore>` on the `Verifier` structure instead of `Vec<CertificateDer<'static>` and reuse that when encountering partial chain errors.

cpu:
Apologies I had queued review comments I hadn't submitted from a few days ago. I think Complexspaces and I overlapped in some feedback as a result.

> One open question I have is about the infallibility of new_with_extra_roots. After seeing how we need to implement this on Windows, I'm wondering if we want to return Result<Self, TlsError> instead so that we can do OS-level logic in the constructor instead of being restricted to infallible operations.

I would be in favour of this :+1:

cpu:
> The Android job failure seems unrelated, as it's already on main.

https://github.com/rustls/rustls-platform-verifier/pull/140

rustls-platform-verifier/src/verification/windows.rs
stormshield-gt: Sure, I've done it

rustls-platform-verifier/src/verification/windows.rs
stormshield-gt: Nice catch, indeed this no longer needed

rustls-platform-verifier/src/verification/windows.rs
stormshield-gt: That's much cleaner, thank you

rustls-platform-verifier/src/verification/windows.rs
stormshield-gt: I've added a comment trying to explicit that fact

rustls-platform-verifier/src/verification/windows.rs
```diff
@@ -521,7 +565,24 @@ impl Verifier {
             .chain(Some(0))
             .collect();
 
-        let cert_chain = store.new_chain_in(&primary_cert, now)?;
+        let mut cert_chain = store.new_chain_in(&primary_cert, now)?;
+
+        // If we have extra roots and the building the chain gave us an error, we try to build a
+        // new one with the extra roots.
+        if !self.extra_roots.is_empty()
```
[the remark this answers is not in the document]
stormshield-gt: nice catch, I've updated the commment

rustls-platform-verifier/src/verification/windows.rs
```diff
@@ -521,7 +565,24 @@ impl Verifier {
...
+            // SAFETY: The pointer is guaranteed to be non-null.
+            && unsafe { *cert_chain.inner.as_ptr() }
```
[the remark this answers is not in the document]
stormshield-gt: That's much cleaner indeed, I've already fix that because @complexspaces raised the same concern

rustls-platform-verifier/src/tests/verification_mock/mod.rs
```diff
@@ -43,7 +43,7 @@ macro_rules! mock_root_test_cases {
                 }
 
                 paste::paste!{
-                    #[cfg(all($target, not(windows), not(target_os = "android")))]
+                    #[cfg(all($target, not(target_os = "android")))]
```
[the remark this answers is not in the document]
stormshield-gt: No you are totally right, I've changed the commit order

rustls-platform-verifier/src/verification/windows.rs
```diff
@@ -34,16 +34,18 @@ use std::{
     ptr::{self, NonNull},
     sync::Arc,
 };
+use windows_sys::Win32::Security::Cryptography::CERT_TRUST_IS_PARTIAL_CHAIN;
```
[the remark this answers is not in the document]
stormshield-gt: That's fixed

stormshield-gt:
Thank you both for the thorough review!

> One open question I have is about the infallibility of new_with_extra_roots. After seeing how we need to implement this on Windows, I'm wondering if we want to return Result<Self, TlsError> instead so that we can do OS-level logic in the constructor instead of being restricted to infallible operations.

I also totally agree this could be better this way. I've implemented a new constructor where I parse the certificates right away.

I now store the certificate engine in the verifier to avoid creating a store each time for the extra roots.

cpu (approved):
This looks good to me within my limited understanding of the `windows_sys` APIs. Thank you @stormshield-gt !

rustls-platform-verifier/src/verification/windows.rs
```diff
@@ -214,6 +212,90 @@ impl Drop for Certificate {
...
+
+impl Drop for CertEngine {
+    fn drop(&mut self) {
+        // SAFETY: The engine pointer is guaranteed to be non-null.
+        unsafe { CertFreeCertificateChainEngine(self.as_ptr()) };
+    }
+}
+
+// SAFETY: We know no other threads is mutating the `CertEngine`, because it would require `unsafe`.
+// Across the FFI, `CertGetCertificateChain` don't mutate it either.
```
complexspaces, line 293: No action needed: I'm not 100% certain of this assertion FWIW, but I think there's enough supporting evidence to go with it. Windows handles are generally thread safe, the `CertFreeCertificateChainEngine` function doesn't say it invalidates the handle, and an engine is really just a set of parameters Windows needs to read.

complexspaces (approved):

rustls-platform-verifier/src/verification/windows.rs
```diff
@@ -214,6 +212,90 @@ impl Drop for Certificate {
...
+unsafe impl Sync for CertEngine {}
+// SAFETY: All methods of `CertEngine`, including `Drop`, are safe to be called from other
+// threads.
```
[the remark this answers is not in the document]
stormshield-gt: Thanks for the extra clarification, I've added to the comment

cpu:
@djc @ctz Do either of you want to give this a pass or should we merge?

rustls-platform-verifier/src/verification/apple.rs
```diff
@@ -72,21 +72,30 @@ impl Verifier {
     /// facilities with the addition of extra root certificates to trust.
     ///
     /// See [Verifier::new] for the external requirements the verifier needs.
-    pub fn new_with_extra_roots(roots: Vec<pki_types::CertificateDer<'static>>) -> Self {
-        Self {
-            extra_roots: roots,
+    pub fn new_with_extra_roots(
+        roots: Vec<pki_types::CertificateDer<'static>>,
+    ) -> Result<Self, TlsError> {
```
ctz, line 77: how about bumping the crate version to `0.5.0` in this commit, so we don't forget the next release is a breaking one?

ctz (approved):

rustls-platform-verifier/src/verification/apple.rs
stormshield-gt: We already bump the version to 0.4 for this reason https://github.com/rustls/rustls-platform-verifier/pull/133/commits/094341ed7c8de588fdeceabdadd397be9cc371c2

cpu: Ahh! Good catch, sorry about missing that. :+1:

stormshield-gt: no problem ;)

ctz: Ah, great!

djc:
No need to block on me; I looked at it but didn't have time to dig into the details of the Windows API.

cpu:
Thanks again for your work on this @stormshield-gt :tada:

PR #136, opened 2024-08-30
Fix and defend wasm code using wasi platform

ctz:
This avoids all the annoyance with the `wasm32-unknown-unknown` platform, and arguably wasi is more typical for a networked application.

This PR is a variation on #122.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -67,12 +67,11 @@ jobs:
           rustup target add x86_64-apple-ios
           cargo clippy-ci --target x86_64-apple-ios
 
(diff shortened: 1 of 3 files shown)


djc (approved):

complexspaces:
I think these changes make a lot of sense given that browser's don't let WASM/JS control what TLS roots are trusted, make raw network connections, etc. Thanks for the proposal :)

It might be worth figuring out how to keep CI happy for `wasm32-unknown-unknown` to make downstream user's conditional compilation easier but that is just a nice to have.

complexspaces (approved):

PR #142, opened 2024-09-18
Add deployment considerations section to README

complexspaces:
This PR adds documentation and more explicit recommendations to the README of the crate to support its preferred use in the Rust app and library ecosystem. This has been a WIP stash for a pretty long time ([since this old Discord conversation](https://discord.com/channels/976380008299917365/1015156984007381033/1141391213736513618)) and I've really been wanting to finish it out for several reasons.

These changes also include noting tradeoffs that `rustls` users might need to consider when doing something different to show supporting evidence.

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,7 +53,7 @@ are primarily about root certificate availability:
 
 | Backend                                         | Updates                         | Roots used                                                                                            | Supports system-local roots  |
 |-------------------------------------------------|---------------------------------|-------------------------------------------------------------------------------------------------------|------------------------------|
-| `rustls-platform-verifier` (non-Linux/BSD)      | Updated by OS                   | System store, with correct (dis)trust decisions from every source available.                          | Yes                          |
+| `rustls-platform-verifier` (non-Linux/BSD)      | Updated by OS                   | System store, with full (dis)trust decisions from every source available.                             | Yes                          |
 | `rustls-native-certs` + `webpki`                | Updated by OS                   | System store, with no (dis)trust decisions. All roots are treated equally regardless of their status. | Yes, with exceptions         |
 | `webpki-roots` + `webpki`                       | Static, manual updates required | Hardcoded Mozilla CA roots, limited support for constrained roots.                                    | No                           |
 
@@ -62,12 +62,22 @@ making connections to TLS servers when running on common operating systems. This
 and better matches the common expectation of apps running on that platform (to use proxies, for example). Otherwise, it becomes your maintenance burden to
 ship updates right away in order to handle increasing numbers of positive and negative trust events in the WebPKI/certificate ecosystem, or risk availability and security concerns.
 
+#### Linux/BSD
+As of the time of writing, `rustls-platform-verifier` on these OSes only loads the trust stores from the OS once upon startup. This is the same behavior as `rustls-native-certs`, but the
+abstraction allows better behavior on the other platforms without extra work for downstreams.
+
+#### Other
+
 Alternatively, there is a clear answer to use static `webpki-roots` in your application instead if you are deploying containerized applications frequently, where root store changes
 will make it to production faster and any possibly used trust root is static by definition.
 
 Even though platform verifiers are sometimes implemented in memory-unsafe languages, it is very unlikely that Rust apps using this library will become a point of weakness.
 This is due to either using a smaller set of servers or just being less exposed then other critical functions of the operating system, default web browser, etc.
-But if you are, for example, fuzzing or scanning all certificates on the open internet, using a 100% Rust option like `webpki` is a better and more secure option.
+But if your activity is identical or close to one of the following examples that process large amounts of untrusted input, a 100% Rust option like `webpki` is a more secure option: 
+- Seeing how many TLS servers `rustls` with a specific configuration can connect to.
+- Harvesting data from various untrusted TLS endpoints exposed on the internet.
+- Extracting info from a known-evil endpoint.
+- Scanning all TLS certificates on the open internet.
 
 `rustls-platform-verifier` is widely deployed by several applications that use the `rustls` stack, such as 1Password, Bitwarden, Signal, and `rustup`, on a wide set of OSes. 
 This means that it has received lots of exposure to edge cases and has real-world experience/expertise invested into it to ensure optimal compatibility and security.
--- a/README.md
+++ b/README.md
@@ -46,6 +46,29 @@ a system CA bundle is unavailable.
 [openssl-probe]: https://github.com/alexcrichton/openssl-probe
 [webpki-roots]: https://github.com/rustls/webpki-roots
 
+## Deployment Considerations
+
+When choosing to use `rustls-platform-verifier` or another trust store option, there are important differences to consider. These
+are primarily about root certificate availability:
+
+| Backend                          | Updates                         | Roots used                                                                                            |
+|----------------------------------|---------------------------------|-------------------------------------------------------------------------------------------------------|
+| OS/platform (non-Linux/BSD)      | Live pushes                     | System store, with correct (dis)trust decisions from every source available.                          |
+| `rustls-native-certs` + `webpki` | Live pushes                     | System store, with no (dis)trust decisions. All roots are treated equally regardless of their status. |
+| `webpki-roots` + `webpki`        | Static, manual updates required | Hardcoded Mozilla CA roots, and all roots are treated equally.                                        |
+
+**In general**: It is the opinion of the `rustls` team and platform verifier maintainers that this is the best default available for client-side libraries and applications
+making connections to TLS servers when running on common operating systems. This is because it gets both live trust information (new roots, explicit markers, and auto-managed CRLs)
+and better matches the common expectation of apps running on that platform (to use proxies, for example). Otherwise, it becomes your maintenance burden to
+ship updates right away in order to handle increasing numbers of positive and negative trust events in the WebPKI/certificate ecosystem, or risk availability and security concerns.
+
+Even though platform verifiers are sometimes implemented in memory-unsafe languages, it is very unlikely that Rust apps using this library will become a point of weakness.
+This is due to either using a smaller set of servers or just being less exposed then other critical functions of the operating system, default web browser, etc.
+But if you are, for example, fuzzing or scanning all certificates on the open internet, using a 100% Rust option like `webpki` is a better and more secure option.
+
+`rustls-platform-verifier` is widely deployed by several applications that use the `rustls` stack, such as 1Password, Bitwarden, Signal, and `rustup`, on a wide set of OSes. 
+This means that it has received lots of exposure to edge cases and has real-world experience/expertise invested into it to ensure optimal compatibility and security.
+
 ## Installation and setup
 On most platforms, no setup should be required beyond adding the dependency via `cargo`:
 ```toml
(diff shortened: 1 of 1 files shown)


djc (approved):
This is great!

I think it would be good to explain the Linux/BSD situation in a little more detail. In particular, while it inherits trust store updates from the OS, currently these are only configured on program startup, so in its current state on Linux/BSD the platform verifier ends up being the same as just using rustls-native-certs directly AFAICT.

Also it might be good to talk about scenarios where the use of webpki-roots could make sense, for example in applications that are (a) deployed frequently in (b) minimal containers.

README.md
```diff
@@ -46,6 +46,29 @@ a system CA bundle is unavailable.
...
+## Deployment Considerations
+
+When choosing to use `rustls-platform-verifier` or another trust store option, there are important differences to consider. These
+are primarily about root certificate availability:
+
+| Backend                          | Updates                         | Roots used                                                                                            |
+|----------------------------------|---------------------------------|-------------------------------------------------------------------------------------------------------|
+| OS/platform (non-Linux/BSD)      | Live pushes                     | System store, with correct (dis)trust decisions from every source available.                          |
+| `rustls-native-certs` + `webpki` | Live pushes                     | System store, with no (dis)trust decisions. All roots are treated equally regardless of their status. |
+| `webpki-roots` + `webpki`        | Static, manual updates required | Hardcoded Mozilla CA roots, and all roots are treated equally.                                        |
```
ctz, line 58: We apply "mozilla applied constraints" via this route, which (at the current time) technically constrains one of the Turkish CAs to issuing for *.tr ([ref](https://bugzilla.mozilla.org/show_bug.cgi?id=1349705))

So I would perhaps reword this as:

```suggestion
| `webpki-roots` + `webpki`        | Static, manual updates required | Hardcoded Mozilla CA roots, limited support for constrained roots.                                        |
```

README.md
```diff
@@ -46,6 +46,29 @@ a system CA bundle is unavailable.
 [openssl-probe]: https://github.com/alexcrichton/openssl-probe
 [webpki-roots]: https://github.com/rustls/webpki-roots
 
+## Deployment Considerations
+
+When choosing to use `rustls-platform-verifier` or another trust store option, there are important differences to consider. These
+are primarily about root certificate availability:
+
+| Backend                          | Updates                         | Roots used                                                                                            |
```
ctz, line 54: I think another thing (either for this table, or another) is whether each option supports locally administered roots (for developer aids like mkcert, or corporate TLS inspection). That is one facet that could make webpki-roots unsuitable.

ctz (commented):
Thanks for writing this down. Once we land this, I think we can delete https://github.com/rustls/rustls-native-certs/?tab=readme-ov-file#should-i-use-this-or-webpki-roots and refer to this as a single source.

README.md
```diff
@@ -46,6 +46,29 @@ a system CA bundle is unavailable.
 [openssl-probe]: https://github.com/alexcrichton/openssl-probe
 [webpki-roots]: https://github.com/rustls/webpki-roots
 
+## Deployment Considerations
+
+When choosing to use `rustls-platform-verifier` or another trust store option, there are important differences to consider. These
```
cpu, line 51: editorial nit: I think it reads a little more naturally to say "These differences are ..." instead of "These are"

README.md
```diff
@@ -46,6 +46,29 @@ a system CA bundle is unavailable.
...
+are primarily about root certificate availability:
+
+| Backend                          | Updates                         | Roots used                                                                                            |
+|----------------------------------|---------------------------------|-------------------------------------------------------------------------------------------------------|
+| OS/platform (non-Linux/BSD)      | Live pushes                     | System store, with correct (dis)trust decisions from every source available.                          |
```
cpu, line 56: Should this `Backend` column be labelled `rustls-platform-verifier (non-Linux/BSD)` to match the other backend options that are phrased in terms of specific crate combinations? I think if you read the full README it's clear that `rustls-platform-verifier` is using the OS/platform but it might be clearer to have the crate name here.

cpu: Separately I also wonder about the word "correct" for describing the (dis)trust decisions. The other crates are not necessarily "incorrect".

Would a word like "full" or phrasing like "most accurate" be better?

README.md
```diff
@@ -46,6 +46,29 @@ a system CA bundle is unavailable.
 [openssl-probe]: https://github.com/alexcrichton/openssl-probe
 [webpki-roots]: https://github.com/rustls/webpki-roots
 
+## Deployment Considerations
+
+When choosing to use `rustls-platform-verifier` or another trust store option, there are important differences to consider. These
+are primarily about root certificate availability:
+
+| Backend                          | Updates                         | Roots used                                                                                            |
+|----------------------------------|---------------------------------|-------------------------------------------------------------------------------------------------------|
```
cpu, line 55: I think instead of "live pushes" it might be clearer to say "OS managed", or "Updated by OS"?

README.md
```diff
@@ -46,6 +46,29 @@ a system CA bundle is unavailable.
...
+| OS/platform (non-Linux/BSD)      | Live pushes                     | System store, with correct (dis)trust decisions from every source available.                          |
+| `rustls-native-certs` + `webpki` | Live pushes                     | System store, with no (dis)trust decisions. All roots are treated equally regardless of their status. |
+| `webpki-roots` + `webpki`        | Static, manual updates required | Hardcoded Mozilla CA roots, and all roots are treated equally.                                        |
+
+**In general**: It is the opinion of the `rustls` team and platform verifier maintainers that this is the best default available for client-side libraries and applications
```
cpu, line 60: I think "this" is a little bit ambiguous here. Maybe:
"It is the opinion of the `rustls` team and the `rustls-platform-verifier` maintainers that this crate is the best default ..." ?

README.md
```diff
@@ -46,6 +46,29 @@ a system CA bundle is unavailable.
...
+making connections to TLS servers when running on common operating systems. This is because it gets both live trust information (new roots, explicit markers, and auto-managed CRLs)
+and better matches the common expectation of apps running on that platform (to use proxies, for example). Otherwise, it becomes your maintenance burden to
+ship updates right away in order to handle increasing numbers of positive and negative trust events in the WebPKI/certificate ecosystem, or risk availability and security concerns.
+
+Even though platform verifiers are sometimes implemented in memory-unsafe languages, it is very unlikely that Rust apps using this library will become a point of weakness.
+This is due to either using a smaller set of servers or just being less exposed then other critical functions of the operating system, default web browser, etc.
```
cpu, line 66: I'm not sure I follow the "smaller set of servers" part. Can you expand on this thought or rephrase?

README.md
```diff
@@ -46,6 +46,29 @@ a system CA bundle is unavailable.
...
+But if you are, for example, fuzzing or scanning all certificates on the open internet, using a 100% Rust option like `webpki` is a better and more secure option.
```
cpu, line 67: I'm not sure I follow the fuzzing angle. I would expect anyone fuzzing is doing so with the intent to find bugs, and so switching to a backend that's less likely to have bugs seems counter-productive. Maybe strike that part of the wording and emphasize a use-case where you're dealing with a large number of untrusted inputs like the "scan the internet" usecase mentioned?

cpu (approved):
This is great, thank you! I had a handful of comments but don't consider any of them blocking. Feel free to disregard if you think I'm barking up the wrong tree on any of my points.

README.md
complexspaces: I think I might have used the word "fuzzing" too ambiguously here. I wasn't trying to refer to the verifier backend but an activity where, like you guessed, you are going to be processing a lot of untrusted TLS data and/or certificates for an experiment. For example:
- Seeing how many TLS servers `rustls` with a specific configuration can connect to
- Harvesting data from various untrusted TLS endpoints exposed on the internet
- Extracting info from a known-evil endpoint (C2 server, etc)

I can try and emphasize a use case but I did want it to be a generic "dont go poking known-evil things with this."

README.md
complexspaces: I don't have that much more to expand with on this point. This paragraph was an attempt to "add some meat" to Joe's original comment on the Discord last year, which roughly summarizes to: the odds that it is the first thing to explode or be exploited on a user's whole computer by touching the network is extremely low. For example this library is unlikely to be used to build a browser (although that would be cool!) and its not an OS networking stack.

README.md
complexspaces: Is it not incorrect to only apply half or none of a distrust decision the rest of the platform is following?

complexspaces:
I addressed most of the first set of feedback. I apologize for taking longer then expected, time slipped away 😅.

The things still pending are:
- Documentation on BSD/Linux, requested by @djc 
- Correct vs full wording for distrust decisions
-  Remaining section clarifications from @cpu

djc (approved):
This looks good to me, thanks for the updates!

README.md
cpu: I don't think it's fair to describe webpki-roots or platform-verifier as incorrect, no. I would still prefer different wording here. There's nuance to the discussion that I think falls away with the binary correct/incorrect.

README.md
cpu: > I wasn't trying to refer to the verifier backend but an activity where, like you guessed, you are going to be processing a lot of untrusted TLS data and/or certificates for an experiment. For example:

That makes sense, but the examples you list here all seem close to the category of "scanning all certificates on the open internet". Does having the "fuzzing" part of the text add much with respect to this thought or is it too ambiguous to be clear?

I still lean towards thinking it would be clearer without but I don't feel very strongly.

cpu (approved):
Thanks! Updates look good. There are still a couple open discussion threads but whether or not they're resolved I think this is a great improvement. Happy for it to land as-is if you disagree with my remaining items.

complexspaces:
I believe I've now addressed all the [remaining feedback](https://github.com/rustls/rustls-platform-verifier/pull/142#issuecomment-2384937062) from today. Please take a final look when available.

README.md
complexspaces: That's valid! I probably took to strong of an approach here, and would prefer to leave it in a state where people who aren't satisfied with the description can come as more specific questions to fit their needs instead of making a blanket assumption. I've changed the wording to "full" per your suggestion.

README.md
complexspaces: > Does having the "fuzzing" part of the text add much with respect to this thought or is it too ambiguous to be clear?

With a pair of fresh eyes I think so, yes. After reading the existing text and comparing it to how I wrote it in this thread the problem became more apparent. I think its best to remove it and just imply "fuzz-like" behavior from the listed examples instead that readers can pick up.

Please have a look at the section again, if you don't mind. I rewrote it to try and be more clear about what constitutes an untrusted use case, borrowed some of your wording, and formatted it bullets to be easier to read/scan.

cpu: The revisions look great, thank you!

cpu (approved):
Thanks again for taking all of our feedback into consideration.

complexspaces:
Of course, and I appreciate the thorough review to make sure the first version of this is high-quality.

complexspaces:
> Once we land this, I think we can delete https://github.com/rustls/rustls-native-certs/?tab=readme-ov-file#should-i-use-this-or-webpki-roots and refer to this as a single source.

I've started that work in [this PR](https://github.com/rustls/rustls-native-certs/pull/140).

djc:
I think it's also time for a release in this crate, right?

PR #143, opened 2024-09-18
ci: add cargo-semver-checks-action

cpu:
We've used [cargo-semver-checks](https://github.com/obi1kenobi/cargo-semver-checks) with success in many other Rustls repos. It's not perfect but catches a variety of semver breaks that could be missed otherwise.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -246,6 +246,18 @@ jobs:
       - name: Verify release artifact
         run: ./ci/verify_android_release.sh
 
+  semver:
+    name: Check semver compatibility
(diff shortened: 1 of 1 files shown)


ctz (approved):

complexspaces (approved):
Nice! I could have sworn we had the tool here already since I thought about doing this same thing during the RustConf talk on `cargo-semver-checks`.

cpu:
> Nice! I could have sworn we had the tool here already 

same! I only thought to check again after the discussion on the other PR about whether a version bump was needed. It must have fallen through the cracks.

PR #146, opened 2024-10-14
ci: check android on ubuntu-22.04

ctz:


ctz:
(I guess we could do this to resolve the immediate problem, but it's probably preferable to adjust the android steps to run on 24.04?)

complexspaces:
See my comment [here](https://github.com/rustls/rustls-platform-verifier/pull/144#issuecomment-2412594731), but I think we have a path to staying on easily `ubuntu-latest`.

Issue #4, opened 2022-11-15
Add iOS integration testing to CI


complexspaces:
To complete test coverage for the native platforms (WASM testing will need handled on its own), we need to add iOS testing inside of CI. This can be done via the simulator included with XCode on the GitHub actions runners. It should run the exact same tests that the Android FFI testing layer currently uses.

Past work in this area includes:
- [libc's iOS testing](https://github.com/rust-lang/libc/tree/master/ci/ios) (This may be partially defunct as-is, it does not look like its being actively used)
- [security-framework's iOS test bindings](https://github.com/kornelski/rust-security-framework/tree/main/iostest) (This is more involved than the `libc` approach, as it requires an XCode project, etc.)

PR #149, opened 2024-10-17
Run tests on iOS using Mac Catalyst and expand Apple platforms

complexspaces:
This PR accomplishes two things:
- Compiling and running all of the tests on an iOS target, by utilizing Catalyst targets as pointed out in [today's 1.82.0 release notes](https://blog.rust-lang.org/2024/10/17/Rust-1.82.0.html). 
    - This is not quite as complete as if we ran these tests in a full simulator, but the environment isn't going to be that different in our use case where we are solely calling `Security.framework` functions. Even in the simulator, a lot of frameworks and kernel functions come "from the macOS host" anyway.
- Expanding the Apple platforms that `rustls-platform-verifier` can be compiled for to all available ones today.
    - AFAICT, all of the `Security.framework` functions we call are available everywhere and have no special OS constraints. All of them make sense to support as well:
    - visionOS is very similar to iOS and has the same networking API availability, etc.
    - watchOS has some network API constraints which restrict what you could perform TLS over, but you could still bring your own transport encryption stack as [per the requirement documentation](https://developer.apple.com/documentation/technotes/tn3135-low-level-networking-on-watchos).

Closes https://github.com/rustls/rustls-platform-verifier/issues/4 (unless anyone thinks we need the simulator explicitly)

### DIFF
--- a/rustls-platform-verifier/Cargo.toml
+++ b/rustls-platform-verifier/Cargo.toml
@@ -36,7 +36,7 @@ jni = { version = "0.19", default-features = false, optional = true } # Only use
 once_cell = "1.9"
 paste = { version = "1.0", default-features = false, optional = true } # Only used when `ffi-testing` feature is enabled
 
-[target.'cfg(all(unix, not(target_os = "android"), not(target_os = "macos"), not(target_os = "ios"), not(target_os = "tvos"), not(target_arch = "wasm32")))'.dependencies]
+[target.'cfg(all(unix, not(target_os = "android"), not(target_vendor = "apple"), not(target_arch = "wasm32")))'.dependencies]
 rustls-native-certs = "0.7"
 webpki = { package = "rustls-webpki", version = "0.102", default-features = false }
 
@@ -54,7 +54,7 @@ webpki-root-certs = "0.26"
 [target.'cfg(target_os = "freebsd")'.dev-dependencies]
 webpki-root-certs = "0.26"
 
-[target.'cfg(any(target_os = "macos", target_os = "ios", target_os = "tvos"))'.dependencies]
+[target.'cfg(any(target_vendor = "apple"))'.dependencies]
 core-foundation = "0.9"
 core-foundation-sys = "0.8"
 security-framework = { version = "2.10", features = ["OSX_10_14"] }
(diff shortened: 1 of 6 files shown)


complexspaces:
This PR is blocked on GitHub updating their runner image to move from Rust 1.81 to 1.82 for a "stable" tier 2 Catalyst target.

rustls-platform-verifier/Cargo.toml
```diff
@@ -36,7 +36,7 @@ jni = { version = "0.19", default-features = false, optional = true } # Only use
 once_cell = "1.9"
 paste = { version = "1.0", default-features = false, optional = true } # Only used when `ffi-testing` feature is enabled
 
-[target.'cfg(all(unix, not(target_os = "android"), not(target_os = "macos"), not(target_os = "ios"), not(target_os = "tvos"), not(target_arch = "wasm32")))'.dependencies]
+[target.'cfg(all(unix, not(target_os = "android"), not(target_vendor = "apple"), not(target_arch = "wasm32")))'.dependencies]
```
djc, line 39: Nit: feels weird to have a first commit that adds `target_os = "ios"` only to then have a commit that replace the Apple `target_os` attributes with `target_vendor = "apple"`, maybe we should do it the other way around and replace with `target_vendor` first, then add specific `ios` stuff where needed?

complexspaces: I originally had this all in one commit but decided to split it out 😅. I can flip the order since you think its better that way.

djc: Yeah, I usually like to minimize within-PR churn between commits.

complexspaces: I've force pushed a re-arrangement of the commits so the Catalyst testing builds ontop of the `target_vendor = "apple"` base instead. No rush to re-review though since it will probably be next week when GitHub bumps their Rust versions.

cpu (approved):
LGTM once the runners update and CI passes. Very nice to get this coverage :rocket:

djc (approved):

complexspaces:
The update finally rolled out so this one was ready 👍

PR #150, opened 2024-10-25
Use extension trait to simplify config

djc:


rustls-platform-verifier/src/lib.rs
```diff
@@ -88,3 +87,28 @@ pub fn tls_config_with_provider(
...
+#[cfg(test)]
+mod setup_tests {
+    use super::*;
+
+    #[test]
+    fn extension_trait() {
+        ClientConfig::builder()
+            .with_platform_verifier()
+            .with_no_client_auth();
+    }
```
ctz, line 113: Maybe put this in a doctest?

ctz (approved):
Seems like a big improvement!

rustls-platform-verifier/src/lib.rs
djc: Ohh, good idea.

djc:
Another idea: we could potentially add another trait implemented directly for `ClientConfig`, such that you could do `ClientConfig::with_platform_verifier()` and `ClientConfig::with_platform_verifier_and_provider(provider)`.

Worth doing over the current approach of `rustls_platform_verifier::tls_config()`?

cpu (approved):
Nice!

cpu:
> Worth doing over the current approach of rustls_platform_verifier::tls_config()?

Your proposal seems a bit tidier to me. I'd be :+1:

complexspaces:
If we don't remove `tls_config`, maybe we should reimplement it in terms of the new extension trait internally for better coverage?

djc:
Implemented a "maximalist" approach with two new traits, that deprecates the old `tls_config*()` API.

complexspaces (approved):

rustls-platform-verifier/src/lib.rs
```diff
@@ -49,17 +51,14 @@ pub use tests::ffi::*;
...
 ///
-/// If you require more control over the rustls `ClientConfig`, you can
-/// instantiate a [Verifier] with [Verifier::default] and then use it
-/// with [`DangerousClientConfigBuilder::with_custom_certificate_verifier`][rustls::client::danger::DangerousClientConfigBuilder::with_custom_certificate_verifier].
+/// If you require more control over the rustls [`ClientConfig`], you can import the
+/// [`BuilderVerifierExt`] trait and call `.with_platform_verifier()` on the [`ConfigBuilder`].
 ///
 /// Refer to the crate level documentation to see what platforms
 /// are currently supported.
+#[deprecated(since = "0.4.0", note = "use the `ConfigVerifierExt` instead")]
```
cpu, line 59: Would you mind updating the deprecated functions described in the README's [Installation and setup section](https://github.com/rustls/rustls-platform-verifier/blob/6927b45f354b780f4a5409f4c070cd246c9703b1/README.md?plain=1#L85-L115)? I think there's both a `tls_config()` and a `tls_config_with_provider()` invocation.

cpu (approved):

djc:
Updated the README and also used the new `ConfigBuilder::crypto_provider()` API to make sure we use the matching crypto provider.

PR #151, opened 2024-11-09
Update jni to 0.21

Cyannide:
Rust JNI 0.21 introduced some breaking changes.

The main reason for needing JNI 0.21 here is that android-activity 0.6.0 is based on it and the JNIEnv references are incompatible.

#63 and #22


rustls-platform-verifier/src/verification/android.rs
```diff
@@ -265,12 +264,17 @@ fn extract_result_info(env: &JNIEnv<'_>, result: JObject<'_>) -> (VerifierStatus
...
+    let msg: Option<String> = if let Ok(msg) = msg {
+        if let Ok(msg) = env.get_string((&msg).into()) {
+            Some(msg.into())
+        } else {
+            None
+        }
+    } else {
+        None
+    };
+    (status, msg)
```
djc, line 277: Nit: maybe we could do it more like this?

```diff
--- a/rustls-platform-verifier/src/verification/android.rs
+++ b/rustls-platform-verifier/src/verification/android.rs
@@ -264,16 +264,10 @@ fn extract_result_info(
     // Extract the `String?`.
     let msg = env
         .get_field(result, "message", "Ljava/lang/String;")
-        .and_then(|m| m.l());
-    let msg: Option<String> = if let Ok(msg) = msg {
-        if let Ok(msg) = env.get_string((&msg).into()) {
-            Some(msg.into())
-        } else {
-            None
-        }
-    } else {
-        None
-    };
+        .and_then(|msg| m.l().ok())
+        .and_then(|msg| env.get_string(&msg).ok())
+        .map(String::from);
+
     (status, msg)
 }
```

djc (approved):
Thanks for picking this up! Could you squash all the JNI-related commits into one so it's cleanly separate from the Let's Encrypt certificate update?

rustls-platform-verifier/src/verification/android.rs
Cyannide: I tried to do that in the chain, but eventually the borrow checker won't let it pass.

djc: Oh, because we need to store some intermediate state on the stack? We can still probably use the same combinators even if we have to bind one of the intermediates?

Cyannide:
> Thanks for picking this up! Could you squash all the JNI-related commits into one so it's cleanly separate from the Let's Encrypt certificate update?

Yep. Will get that done this morning.

cpu:
@Cyannide Thanks for picking this up.

@complexspaces Do you think https://github.com/jni-rs/jni-rs/pull/513 is a blocker for this?

rustls-platform-verifier/src/verification/android.rs
Cyannide: Yeah I got it worked out...had to use an unwrap() but it's ok since the value is guaranteed to be Some from previous step.

Cyannide: Well nvm...it won't work because None does get passed through at that point. I'll change it again.

cpu (approved):
Thanks again. 

I think there's a couple `clippy`/`rustfmt` nits to sort out, but overall this seems :+1: It also looks like the diff in this branch ended up a little bit less involved than in #63, nice to see that.

Beyond getting CI passing I think we should wait for Complexspaces to weigh in w.r.t the upstream window-sys dep before merging.

rustls-platform-verifier/src/android.rs
```diff
@@ -74,10 +74,10 @@ fn global() -> &'static Global {
 /// nothing else in your application needs access the Android runtime.
 ///
 /// Initialization must be done before any verification is attempted.
-pub fn init_hosted(env: &JNIEnv, context: JObject) -> Result<(), JNIError> {
+pub fn init_hosted(env: &mut JNIEnv, context: JObject) -> Result<(), JNIError> {
```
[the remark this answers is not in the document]
Cyannide: You bet. I'll add that as soon as I get the build to come out clean again. Thanks for the input!

complexspaces:
> @complexspaces Do you think https://github.com/jni-rs/jni-rs/pull/513 is a blocker for this?

It _used_ to be a blocker (IMO) until it became clear that `jni` is essentially abandoned, therefore removing the better path of just holding out ~N months and solving the root cause. However now that a legitimate use case for the upgrade has come up (needing compatible types between crates), I don't see a reason not to do this. Additionally while it does cause lockfile bloat, its not nearly as bad as other cases because almost no one is using the `jni` crate not-for-Android. You won't get two `windows-sys` in your active build, the ancient `0.48` copy this brings in will just sit. 

If someone is writing a Java extension that compiles on Windows, and maybe uses this crate: sorry 😅?

FWIW what do people think about dropping the Cargo bump commit currently here? It makes it slightly more annoying to do releases because there's no easy diff to open a branch on without that. We don't guarantee the `main` branch will be properly versioned for anyone using a `git` dependency.

complexspaces:
I'd love to see a maintained version of `jni` this crate could use, but I don't have the time or expertise to make that happen 😔

djc:
> FWIW what do people think about dropping the Cargo bump commit currently here? It makes it slightly more annoying to do releases because there's no easy diff to open a branch on without that. We don't guarantee the `main` branch will be properly versioned for anyone using a `git` dependency.

Not sure I entirely understand the annoyance here? In other rustls projects we like to bump proactively in PRs that make incompatible changes, as a way of recording that incompatible changes have been made. It can be pretty hard to establish after the fact whether incompatible changes have been made once a substantial amount of changes have been made...

Cyannide:
> I'd love to see a maintained version of `jni` this crate could use, but I don't have the time or expertise to make that happen 😔

I'll take a look at the Windows side of things tomorrow to see what the issues there are to overcome. I'm also of the opinion that if someone is honestly building something on Windows that requires the JNI dependency, sorry.

complexspaces:
FWIW, If I'm understanding you correctly, I don't think its really about overcoming issues. The `jni` developers have been presented two options, and responded to neither as of writing this:
- Update `windows-sys` to a reasonable version.
- Use `windows-bindgen` instead to call the very few functions it needs manually to avoid this problem in the future (I offered to implement it for them).

cpu:
> I'm happy to give this a proper review/lookover tomorrow or Wednesday.

gentle bump.

If it's going to take you a little bit to find time to review the meat of this branch maybe we could pull out the CI fix separately so I don't get emails about the cron build failing? :bow:

complexspaces:
> gentle bump.

Thanks! This looks good to go 👍. 

I had looked over this last week but I just did it offhand and not in an actual GitHub review.

PR #152, opened 2024-12-02
Bump MSRV & update dependencies

djc:


complexspaces:
Thanks for the PR! I am a little too busy this week but I'm leaving acknowledgement that I will have time to stamp this by Saturday.

complexspaces (approved):
Thanks! This is a lot cleaner without all the nested matching and error handling.

The PII log safety of the errors still look good too.

PR #154, opened 2024-12-06
ci: add some missing persist-credentials

cpu:
Most of the checkout action usages in `ci.yml` already specified `persist-credentials: false`, but a few were missing it. This [is flagged](https://woodruffw.github.io/zizmor/audits/#artipacked) when running Zizmor 0.7.0 on the repo's CI config. As mentioned in the description of this finding:

> By default, using [actions/checkout](https://github.com/actions/checkout) causes a credential to be persisted in the checked-out repo's .git/config, so that subsequent git operations can be authenticated.
> 
> Subsequent steps may accidentally publicly persist .git/config, e.g. by including it in a publicly accessible artifact via [actions/upload-artifact](https://github.com/actions/upload-artifact).
> 
> However, even without this, persisting the credential in the .git/config is non-ideal unless actually needed.

We don't need it, so turn it off consistently :-)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -147,6 +147,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
+        with:
+          persist-credentials: false
 
       # Turn on Linux KVM features/support for faster Android emulation.
       # References:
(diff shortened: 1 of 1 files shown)


cpu:
CI won't pass until https://github.com/rustls/rustls-platform-verifier/pull/152 lands.

ctz (approved):

djc (approved):

complexspaces (approved):
Thanks for taking the time to go through some CI auditing 👀.

complexspaces:
I just reviewed and merged #152, so if you rebase this should be good to go as well.

PR #155, opened 2025-01-31
Update the crate version in README

1hakusai1:
The examples in README can not be compiled because `ConfigVerifierExt` and `BuilderVerifierExt` don't exist in v0.3.
It should be updated to the latest version.

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,7 +85,7 @@ This means that it has received lots of exposure to edge cases and has real-worl
 ## Installation and setup
 On most platforms, no setup should be required beyond adding the dependency via `cargo`:
 ```toml
-rustls-platform-verifier = "0.3"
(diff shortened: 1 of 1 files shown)


1hakusai1:
The tests for Android has failed, but it seems not to be related to this change. (This test has failed multiple times in the past.)

djc (approved):

djc:
Thanks!

complexspaces (approved):
Thanks for this! I'm trying to take a look at the test failure which I suspect is just the real world data having expired too long in #156.

complexspaces:
It looks like #156 might take a while to figure out, so given the simplicity of this PR and how 19/20 tests passed I am going to override merge this so the README on the main GitHub page is visibly correct.

PR #156, opened 2025-01-31
Update CI for January 2025

complexspaces:
This PR does two things for CI maintenance:
- Updates real world test certs via running `update_valid_ee_certs.rs`
    - The new `letsencrypt.org` certificate is issued by the E6 intermediate instead of E5. To match I updated that chain's intermediate locally by downloading it from https://letsencrypt.org/certificates/.
- Moves to a non-deprecated Nix cache action, per the current GHA deprecation warning I saw on #155.


complexspaces:
Either I ran the update script wrong or something else bitrotted with it and the certificate update went south. The new one seems completely wrong according to CI.

complexspaces:
This is a new failure: `cargo test` passes locally for me (macOS 15.3) but in CI we are now getting this error I haven't seen before:
```
---- tests::verification_real_world::tests::letsencrypt stdout ----
thread 'tests::verification_real_world::tests::letsencrypt' panicked at rustls-platform-verifier/src/tests/mod.rs:51:9:
assertion `left == right` failed
  left: Err(InvalidCertificate(Other(OtherError("“letsencrypt.org” certificate is not standards compliant: -67825"))))
 right: Ok(())
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

The error code in the headers corresponds to a generic verification failure, so that error message might be all we have:
```
errSecVerifyActionFailed                 = -67825,    /* A verify action has failed. */
```

I'm not sure what would be causing this. If I convert the PEM given to me by Firefox on `letsencrypt.org`, it matches byte-by-byte compared to what I've checked in and pushed. I think we can rule out it being valid too long because:
- The end entity is valid for the usual 90 days
- The updated E6 intermediate doesn't have any longer of a life then the previous E5 one.

complexspaces:
Someone elsewhere said that they saw this on a production macOS system and it was fixed eventually by updating macOS. I tried running the tests in macOS 15 (`macos-latest` is still at 14) and 13 runners but the same issue occurs.

complexspaces:
As an update here, I was able to finally get the test failures reproducing locally 😦. Out of ideas I just tried updating the LE leaf certificate again since it renewed recently and that seemed to have just fixed the "standards compliance" issue macOS was upset about.

As CI is now passing: @djc @cpu do you mind stamping this to get #158 unblocked?

cpu:
> Out of ideas I just tried updating the LE leaf certificate again since it renewed recently and that seemed to have just fixed the "standards compliance" issue macOS was upset about.

So weird! :confused:

cpu (approved):
Thanks for fixing this

ctz (approved):

PR #159, opened 2025-02-12
Refactor Android init methods for more flexibility

mcginty:
After integrating `rustls-platform-verifier` into a codebase that uses Android, I wrote this PR to propose a more flexible initialization API. Specifically, it would be nice to, rather than passing in a context `JObject` that isn't `Copy`/`Clone`, to instead allow people to pass in a `GlobalRef` directly, which is what already happens behind the scenes in `init_hosted`.

I also modified the names of the existing methods as well to try to be more clear about what each of the 3 init options offer. I also kept the old names and marked them as deprecated to allow for backwards-compatibility.
| Old | New |
| --- | --- |
| `fn init_hosted(env: &mut JNIEnv, context: JObject) -> Result<(), JNIError>` | `fn init_with_env(env: &mut JNIEnv, context: JObject) -> Result<(), JNIError>` |
| `fn init_external(runtime: &'static dyn Runtime)`                            | `fn init_with_runtime(runtime: &'static dyn Runtime)`                          |
|                                                                         | `fn init_with_refs(java_vm: JavaVM, context: GlobalRef, loader: GlobalRef)`    |

### DIFF
--- a/rustls-platform-verifier/src/android.rs
+++ b/rustls-platform-verifier/src/android.rs
@@ -1,4 +1,23 @@
-//! A small wrapper over interacting with the JNI in a type-safe way.
+//! On Android, initialization must be done before any verification is attempted.
+//!
+//! <div class="warning">
+//! Some manual setup is required, outside of cargo, to use this crate on Android. In order to use
+//! Android’s certificate verifier, the crate needs to call into the JVM. A small Kotlin component
+//! must be included in your app’s build to support `rustls-platform-verifier`.
+//!
+//! See the [crate's Android section][crate#android] for more details.
+//! </div>
+//!
+//! # Examples
+//!
+//! ```
+//! // A typical entrypoint signature for obtaining the necessary pointers
+//! pub fn android_init(raw_env: *mut c_void, raw_context: *mut c_void) -> Result<(), jni::errors::Error> {
+//!     let mut env = unsafe { JNIEnv::from_raw(raw_env as *mut jni::sys::JNIEnv).unwrap() };
+//!     let context = unsafe { JObject::from_raw(raw_context as jni::sys::jobject) };
+//!     rustls_platform_verifier::android::init_with_env(&mut env, context)?;
+//! }
+//! ```
 
 use jni::errors::Error as JNIError;
 use jni::objects::{GlobalRef, JClass, JObject, JValue};
--- a/rustls-platform-verifier/src/android.rs
+++ b/rustls-platform-verifier/src/android.rs
@@ -1,9 +1,9 @@
 //! On Android, initialization must be done before any verification is attempted.
 //!
 //! <div class="warning">
-//! Some manual setup is required, outside of cargo, to use this crate on Android. In order to use
+//! Some manual setup is required outside of cargo to use this crate on Android. In order to use
 //! Android’s certificate verifier, the crate needs to call into the JVM. A small Kotlin component
-//! must be included in your app’s build to support `rustls-platform-verifier`.
+//! must be included in your app’s build to support rustls-platform-verifier.
 //!
 //! See the [crate's Android section][crate#android] for more details.
 //! </div>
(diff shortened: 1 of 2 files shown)


mcginty:
@djc good call, I like that plan better too. It also makes me think I can do a better job at documenting the differences in the rustdocs, so I'll take care of that and report back when they're broken into independent commits.

mcginty:
Alright, done! How does this rendition look?

rustls-platform-verifier/src/android.rs
```diff
@@ -68,33 +68,35 @@ fn global() -> &'static Global {
...
+}
+
 /// Initializes and stores the required context for the Android platform.
 ///
-/// This method utilizes an existing Android runtime environment and set anything
-/// else up on its own. This is useful if your application already interacts with
-/// the runtime and has pre-existing handles.
+/// This method utilizes an existing Android runtime environment and set anything else up on its
+/// own. This is useful if your application already interacts with the runtime and has pre-existing
+/// handles.
```
djc: Tiny nit: avoid these changes in this commit, leave them for the next commit that changes this function?

djc (approved):
This is much better!

My review is by no means sufficient for this code, but this looks okay to me.

rustls-platform-verifier/src/android.rs
```diff
@@ -68,41 +87,76 @@ fn global() -> &'static Global {
...
-/// Initializes and stores the required context for the Android platform.
+/// Initialize given a typical Android NDK `JNIEnv` and `jobject` context.
 ///
-/// This method will setup and store an environment locally. This is useful if
-/// nothing else in your application needs access the Android runtime.
-///
-/// Initialization must be done before any verification is attempted.
-pub fn init_hosted(env: &mut JNIEnv, context: JObject) -> Result<(), JNIError> {
+/// This method will setup and store an environment locally. This is useful if nothing else in your
+/// application needs access the Android runtime.
```
complexspaces, line 93: This is my typo originally: "needs access the Android runtime", but if you wouldn't mind squashing it in your existing rustdoc commit I'd be appreciated 😄.

complexspaces (approved):
This looks good, thanks for the addition. I built the docs locally to confirm the warning text shows up correctly as well:
image

rustls-platform-verifier/src/android.rs
```diff
@@ -68,33 +68,35 @@ fn global() -> &'static Global {
         .expect("Expect rustls-platform-verifier to be initialized")
 }
 
-/// Initializes and stores the required context for the Android platform.
-///
-/// This method will setup and store an environment locally. This is useful if
-/// nothing else in your application needs access the Android runtime.
+/// Initialize given a typical Android NDK `JNIEnv` and `jobject` context.
```
cpu, line 90: nit: `jobject` -> `JObject` in this docstring to match the casing of the type being ref'd.

rustls-platform-verifier/src/android.rs
```diff
@@ -68,33 +68,35 @@ fn global() -> &'static Global {
...
+pub fn init_hosted(env: &mut JNIEnv, context: JObject) -> Result<(), JNIError> {
+    init_with_env(env, context)
+}
+
 /// Initializes and stores the required context for the Android platform.
 ///
-/// This method utilizes an existing Android runtime environment and set anything
-/// else up on its own. This is useful if your application already interacts with
-/// the runtime and has pre-existing handles.
+/// This method utilizes an existing Android runtime environment and set anything else up on its
```
cpu: I think there's a small pre-existing typo being carried forward here too: 

"and set anything else up -> "and set**s** anything else up"

Maybe it's just sufficient to say it uses an existing Android runtime environment and leave it at that?

cpu (approved):
Just a few small nits to note. Nice tidy :-)

rustls-platform-verifier/src/android.rs
mcginty: This rustdoc was rewritten in a subsequent commit, and the typo is no longer there!

mcginty:
Pushed a commit with all the fixes for the typos mentioned in the review and while going through it one more time, realized that the warning block had some unnecessary commas and wasn't rendering markdown backticks as expected, so I just removed them.

complexspaces (approved):

PR #161, opened 2025-02-22
Allow modern range of windows-sys versions

complexspaces:
This PR is based off my past work (and all of its referenced prior art) in https://github.com/1Password/arboard/pull/176/. The goal of this PR is to allow downstream users more flexibility in what version(s) of `windows-sys` cargo resolves into their lockfile.

By allowing two versions in the range, users have a higher chance of being able to only build one version and we appease both the users who want to keep their dependencies stable while not holding back the wider ecosystem and users who want to be on the latest versions all the time.

These changes also include a new CI task to ensure that nothing added to `windows.rs` runs into compatibility issues with `windows-sys` 0.59 (or any future versions we may add).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -83,6 +83,14 @@ jobs:
           rustup target add wasm32-wasip1
           cargo clippy-ci --target wasm32-wasip1
 
+      # Update any flexible dependencies that may have SemVer applicable changes
+      # to ensure the project keeps building with the whole version range. This is:
+      # - windows-sys 0.52, which `rustls-platform-verifier` also should accept at 0.59.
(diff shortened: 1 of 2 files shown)


complexspaces:
I don't believe this puts any meaningful restrictions on improving the Windows verifier code in the future. The only thing that came to mind was better size assertions for `CERT_CHAIN_PARA` but that hasn't become possible with [0.59 either](https://docs.rs/windows-sys/latest/windows_sys/Win32/Security/Cryptography/struct.CERT_CHAIN_PARA.html).

djc (approved):

cpu (approved):
Makes sense to me 👍 thanks for linking in the broader context.

PR #162, opened 2025-03-09
Fix audit

djc:
This is technically semver-incompatible because it removes the `paste` feature. I think we can ignore this because it's only used for testing and so it seems unlikely that someone in the wild depends on it.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -13,9 +13,9 @@ dependencies = [
 
 [[package]]
 name = "android_log-sys"
-version = "0.3.1"
+version = "0.3.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5ecc8056bf6ab9892dcd53216c83d1597487d7dacac16c8df6b877d127df9937"
+checksum = "84521a3cf562bc62942e294181d9eef17eb38ceb8c68677bc49f144e4c3d4f8d"
 
 [[package]]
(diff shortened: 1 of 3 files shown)


ctz (approved):

Cargo.lock
```diff
@@ -278,15 +271,14 @@ dependencies = [
...
- "paste",
  "rustls",
  "rustls-native-certs",
  "rustls-platform-verifier-android",
  "rustls-webpki",
  "security-framework",
  "security-framework-sys",
  "webpki-root-certs",
- "windows-sys 0.52.0",
+ "windows-sys 0.59.0",
```
complexspaces, line 281: Suggestion: I believe this line needs reverted or otherwise we won't be getting CI coverage over both 0.52 and 0.59 to match the semver range specified in the manifest.

complexspaces (changes requested):
> I think we can ignore this because it's only used for testing and so it seems unlikely that someone in the wild depends on it.

I agree with this. The ffi-testing feature is explicitly documented as unstable/internal so it's not covered by the version's compatibility guarantees.

Cargo.lock
djc: Ah, good catch. Reverted.

complexspaces (approved):
Sorry for the review time gap, LGTM.

Issue #67, opened 2024-02-04
Android - Failed to resolve: rustls:rustls-platform-verifier


Ghamza-Jd:
Hey 👋

I'm trying to set this package for android. I've followed the instructions, but I'm using `gradle.kts`.

Here's my `build.gradle.kts`:

```kotlin
import com.android.aaptcompiler.android.isTruthy
import groovy.json.JsonSlurper
import java.net.URI
import java.util.Locale

plugins {
    id("com.android.library")
    id("org.jetbrains.kotlin.android")
}

repositories {
    google()
    mavenCentral()
    gradlePluginPortal()
    maven {
        url = findRustlsPlatformVerifierProject()
        metadataSources.artifact()
    }
}

android {
    namespace = "com.ghamza.jarust"
    compileSdk = GradleConfigs.compileSdk
    ndkVersion = GradleConfigs.ndkVersion

    defaultConfig {
        minSdk = GradleConfigs.minSdk
        consumerProguardFiles("consumer-rules.pro")
    }

    compileOptions {
        sourceCompatibility = JavaVersion.VERSION_1_8
        targetCompatibility = JavaVersion.VERSION_1_8
    }

    kotlinOptions {
        jvmTarget = "1.8"
    }
}

dependencies {
    implementation(libs.androidx.core.ktx)
    implementation(libs.jna)
    implementation("rustls:rustls-platform-verifier:latest.release")
    // also tried rustls:rustls-platform-verifier:0.1.0
}

fun findRustlsPlatformVerifierProject(): URI {
    val dependencyText = providers.exec {
        workingDir = File("../")
        commandLine("cargo", "metadata", "--format-version", "1", "--manifest-path", "./Cargo.toml")
    }
        .standardOutput
        .asText
        .get()
    val dependencyJson = JsonSlurper().parseText(dependencyText) as Map<*, *>
    val packages = dependencyJson["packages"] as List<*>
    val targetPackage = packages.find {
        val pkgMap = it as Map<*, *>
        pkgMap["name"] == "rustls-platform-verifier-android"
    } as Map<*, *>
    val manifestPath = targetPackage["manifest_path"] as String
    return File(file(manifestPath).parent).toURI()
}
```

If it try to log the uri I get `file:/Users/hamzajadid/.cargo/registry/src/index.crates.io-6f17d22bba15001f/rustls-platform-verifier-android-0.1.0/` which seems correct.

[...]

PR #167, opened 2025-04-21
Add Kotlin Gradle script example to Android README section

### LINKED ISSUE #166: Add documentation examples for kotlin format
I'm trying to set up compilation for Android in an app that has all the configs in `.gradle.kts` format and I have no idea how to rewrite the examples in the documentation (in "groovy" syntax) using the alternative kotlin syntax: https://docs.rs/rustls-platform-verifier/latest/rustls_platform_verifier/#gradle-setup

Eg; in regards to this note from the docs:
```
These steps assume you are using .gradle Groovy files because they’re the most common, but 
everything is 100% applicable to Kotlin script (.gradle.kts) configurations too with a few replacements.
```
What do those "few replacements" actually look like?

complexspaces:
This PR adds a documented example of how to use the Android component when your Android app uses Kotlin script files instead of Groovy scripts to handle the Gradle build system. We've now had ~2-3 people ask about this, so I think that justifies giving the information a proper spot instead of linking to a issue comment over and over.

The implementation is based primarily off [this older example](https://github.com/rustls/rustls-platform-verifier/issues/67#issuecomment-1926238111) I posted last year, with a few changes adapted from 1P's current production scripts. I opted to keep it in a collapsed section to avoid making the README any longer to scroll through.

Closes #166.

### DIFF
--- a/README.md
+++ b/README.md
@@ -117,8 +117,8 @@ component must be included in your app's build to support `rustls-platform-verif
 #### Gradle Setup
 
 `rustls-platform-verifier` bundles the required native components in the crate, but the project must be setup to locate them
-automatically and correctly. These steps assume you are using `.gradle` Groovy files because they're the most common, but everything
-is 100% applicable to Kotlin script (`.gradle.kts`) configurations too with a few replacements.
+automatically and correctly. These steps assume you are using `.gradle` Groovy files because they're the most common, but if you are using
+Kotlin scripts (`.gradle.kts`) for configuration instead, an example snippet is included towards the end of this section.
 
 Inside of your project's `build.gradle` file, add the following code and Maven repository definition. If applicable, this should only be the one "app" sub-project that
 will actually be using this crate at runtime. With multiple projects running this, your Gradle configuration performance may degrade.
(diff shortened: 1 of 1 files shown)


complexspaces:
That's perfectly ok, thanks for the heads up. I just wanted to at least give the other maintainers a chance to look it over. 

Does it sound fine to wait another week or so for Daniel (as the other person who's worked on Android stuff) and after that just proceed with merging this? These are just documentation changes and I'm entirely confident in the Kotlin script itself being correct.

README.md
```diff
@@ -142,7 +142,7 @@ repositories {
 String findRustlsPlatformVerifierProject() {
     def dependencyText = providers.exec {
         it.workingDir = new File("../")
-        commandLine("cargo", "metadata", "--format-version", "1", "--manifest-path", "$PATH_TO_DEPENDENT_CRATE/Cargo.toml")
+        commandLine("cargo", "metadata", "--format-version", "1", "--filter-platform", "aarch64-linux-android", "--manifest-path", "$PATH_TO_DEPENDENT_CRATE/Cargo.toml")
```
cpu, line 145: Should the text above be updated to mention the `--filter-platform`? I assume for larger projects it's probably something that can't be fixed to just `aarch64-linux-android` and needs some special consideration (?)

cpu (approved):
> These are just documentation changes and I'm entirely confident in the Kotlin script itself being correct.

Like djc I have to admit my experience here isn't super helpful. This is the only Gradle project I've used and I've written almost zero Kotlin :laughing: 

That said, it seems to match the implementation of the pre-existing Groovy version and since you're confident I think we should merge it without spending any extra time trying to find expert reviews (it is just an example after all).

README.md
complexspaces: I can't think of any reason you wouldn't be able to hardcode this. ARM64 is the largest platform by far in the Android device ecosystem, so every application is going to target at it at minimum (and likely ARMv7 and x86_64 too), making this a safe bet to pick up on the relevant `rustls-platform-verifier` dependency.

PR #168, opened 2025-04-23
Adapt to changing rustls error API

djc:
Depends on:

- https://github.com/rustls/rustls/pull/2426


complexspaces (approved):

complexspaces:
This looks good to me, I think CI is just upset because of versions not yet released, etc which should clear up as the `webpki` -> `rustls` -> ... updates flow further.

cpu (approved):

PR #171, opened 2025-04-25
Eagerly build root store on miscellaneous Unix platforms

djc:


.github/workflows/ci.yml
```diff
@@ -13,7 +13,7 @@ permissions:
   contents: read
 
 env:
-  RUSTFLAGS: -D warnings -F unused_must_use
+  RUSTFLAGS: -D warnings
```
ctz, line 16: Slightly unclear whether this belongs in this commit?

djc: It was causing issues with the `allow(unused)` I added IIRC. Don't think we need the `-F` anyway when we have `-D warnings`?

ctz: True!

cpu:
Is there some larger context/motivation that can be called out in the PR desc?

djc:
> Is there some larger context/motivation that can be called out in the PR desc?

Yes, done.

rustls-platform-verifier/src/verification/others.rs
```diff
@@ -78,18 +78,10 @@ impl Verifier {
...
-                num_extra_roots += 1;
-            }));
-
-        if num_extra_roots > 0 {
-            if ignored > 0 {
-                log::warn!("{ignored} extra CA root certificates were ignored due to errors");
-            } else {
-                log::debug!("Loaded {added} CA root certificates from extra roots");
-            }
+        // While we ignore invalid certificates from the system, we forward errors from
```
djc, line 81: @paolobarbolini added this change. Should help?

paolobarbolini: Perfect. Thanks for the quick fix!

rustls-platform-verifier/src/lib.rs
```diff
@@ -97,16 +52,22 @@ pub trait BuilderVerifierExt {
...
-    fn with_platform_verifier(self) -> ConfigBuilder<ClientConfig, WantsClientCert> {
+    fn with_platform_verifier(
+        self,
+    ) -> Result<ConfigBuilder<ClientConfig, WantsClientCert>, rustls::Error> {
         let provider = self.crypto_provider().clone();
-        self.dangerous()
-            .with_custom_certificate_verifier(Arc::new(Verifier::new().with_provider(provider)))
+        Ok(self
+            .dangerous()
+            .with_custom_certificate_verifier(Arc::new(Verifier::new(provider)?)))
```
complexspaces, line 70: Very small nit: Now that we have a `?` it reads clearer if the fallible part is extracted out of the parentheses soup:
```rust
let provider = self.crypto_provider().clone();
let verifier = Verifier::new(provider)?;
Ok(self.dangerous().with_custom_certificate_verifier(Arc::new(verifier)))
```

rustls-platform-verifier/src/tests/verification_mock/mod.rs
```diff
@@ -94,17 +96,21 @@ const LOCALHOST_IPV6: &str = "::1";
 #[cfg_attr(feature = "ffi-testing", allow(dead_code))]
 pub(super) fn verification_without_mock_root() {
     ensure_global_state();
+    let crypto_provider = CryptoProvider::get_default().unwrap().clone();
```
complexspaces, line 99: Since we need a line to explicitly grab the crypto provider anyway now (here and the other places in the changeset), I think we can refactor `ensure_global_state` into `get_test_provider`:
```rust
fn get_test_provider() -> rustls::CryptoProvider {
    rustls::crypto::ring::default_provider()
}
```
This also aligns with the other changes in the PR which reduce global/implicit state in the actual library too. Thoughts?

rustls-platform-verifier/src/verification/apple.rs
```diff
@@ -49,23 +49,19 @@ pub struct Verifier {
     /// Testing only: The root CA certificate to trust.
     #[cfg(any(test, feature = "ffi-testing", feature = "dbg"))]
     test_only_root_ca_override: Option<SecCertificate>,
-    pub(super) crypto_provider: OnceCell<Arc<CryptoProvider>>,
+    pub(super) crypto_provider: Arc<CryptoProvider>,
```
complexspaces, line 52: I believe this one, and the other fields on different platform providers, no longer need to be `pub(super)`?

complexspaces (changes requested):
This looks wonderful overall! I just have a few small refactor suggestions.

rustls-platform-verifier/src/lib.rs
djc: Rearranged this slightly, although I still inlined the `provider` into the `verifier` setup statement.

rustls-platform-verifier/src/tests/verification_mock/mod.rs
djc: Yes, great idea! Done.

rustls-platform-verifier/src/lib.rs
complexspaces: That looks much better as well, thanks!

PR #176, opened 2025-05-30
Upgrade to webpki-root-certs 1

djc:


Cargo.lock
```diff
@@ -277,7 +277,7 @@ dependencies = [
  "security-framework",
  "security-framework-sys",
  "webpki-root-certs",
- "windows-sys 0.52.0",
+ "windows-sys 0.59.0",
```
complexspaces, line 280: I think this slipped in accidentally under a `cargo update`. It needs to stay on `0.52` so the minimum version testing covers both.

complexspaces (changes requested):

Cargo.lock
djc: Yeah, it keeps flipping for me. Will revert...

complexspaces (approved):
Hurray for more 1.0 deps 🎉

PR #178, opened 2025-06-01
Inline macro gunk in real world test cases

djc:
I think 3 more lines is well worth the much more idiomatic structure.

Can do the mock tests too if you agree this is a good direction.


complexspaces:
I don't think this works as-is because we've removed the `ALL_TEST_CASES` static list. That's currently [relied upon](https://github.com/rustls/rustls-platform-verifier/blob/c72d9dc8b689c40d115d3862fff63d46b01fecef/rustls-platform-verifier/src/tests/ffi.rs#L120-L136) in the Android FFI tests to let a single entrypoint in the Kotlin test framework run every test we define in Rust without having to worry about syncing them. We'd also use it if we ever added "real" iOS testing inside an XCode simulator.

PR #180, opened 2025-07-02
feat: Allow to ignore server name checks.

calavera:
👋 I'm working on [making Rustls the default crypto implementation for Reqwest](https://github.com/seanmonstar/reqwest/pull/2752).

There is a [discussion about adopting this project as the default verifier](https://github.com/seanmonstar/reqwest/issues/2025#issuecomment-2956991896), instead of WebPKI, which is what Reqwest uses by default at the moment. One of the challenges for that adoption is that Reqwest allows to skip server name verifications. [It implements it's own verifier for this feature](https://github.com/seanmonstar/reqwest/blob/master/src/tls.rs#L662-L723).

If we want to adopt this project in Reqwest, I see two options:

1. Use it for the default behavior, and fallback to its own IgnoreHostname verifier when the option is enabled, using WebPKI in that case.
2. Allow this verifier to ignore hostname checks and use it for all cases. What I'm trying to implement in this PR.

I'm not sure if this feature is something you all would even want to have or provide. From my point of view, it'd be cleaner if Reqwest could only use a verifier for all its features.

I'll appreciate your feedback on this change.

### DIFF
--- a/rustls-platform-verifier/src/verification/android.rs
+++ b/rustls-platform-verifier/src/verification/android.rs
@@ -15,6 +15,7 @@ use rustls::{
 
 use super::{log_server_cert, ALLOWED_EKUS};
 use crate::android::{with_context, CachedClass};
+use crate::verification::HostnameVerification;
 
 static CERT_VERIFIER_CLASS: CachedClass =
     CachedClass::new("org/rustls/platformverifier/CertificateVerifier");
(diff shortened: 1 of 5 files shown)


briansmith:
> * Use it for the default behavior, and fallback to its own IgnoreHostname verifier when the option is enabled, using WebPKI in that case.

I think that is the better option. IDK why it bothers doing the rest of the steps using WebPKI since doing any of the steps without doing the hostname verification is pointless from a security standpoint. Maybe there are some non-security reasons. Regardless...

> * Allow this verifier to ignore hostname checks and use it for all cases. What I'm trying to implement in this PR.

This is too much of a footgun for this crate.

calavera:
Thanks for the feedback @briansmith. I'm closing this PR.

PR #182, opened 2025-08-03
Fix docs build for 0.6.0

### LINKED ISSUE #181: docs.rs failed to build rustls-platform-verifier-0.6.0
Doc build failed for version 0.6.0 and cannot be accessed

ctz:
fixes #181

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -299,9 +299,9 @@ jobs:
           persist-credentials: false
 
       - name: Install rust toolchain
-        uses: dtolnay/rust-toolchain@stable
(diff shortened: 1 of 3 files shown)


ctz:
> Is it essential that the docs use OnceCell (still)?

I think so. android.rs is using it extensively, and it comes with some useful public-facing docs.

complexspaces (approved):
Thanks for grabbing this! It looks like the one test is just expiry related so this is still fine to merge.

Once we can bump our MSRV to 1.70 we can drop the `once_cell` dependency everywhere and use the type from `std` instead which should clean up nicely IMO.

Issue #184, opened 2025-08-05
[Android] Don't use attach_current_thread_permanently, leave that decision up to the client


jrose-signal:
The Android implementation works without additional context by storing the JVM in a global and attaching to it as needed. However, it uses `attach_current_thread_permanently` to do so.

https://github.com/rustls/rustls-platform-verifier/blob/eb809982ab2bd0c2d1b1bec6d45e22a66dee1ad9/rustls-platform-verifier/src/android.rs#L55-L61

This can affect and be affected by any other uses of the JVM from this thread:

- If the thread was originally a Java thread, basically nothing happens.
- If the thread was already attached manually, basically nothing happens.
- If the thread was already attached *as a daemon,* nothing happens, but it's unclear to me whether the JVM can still *shut down* while the thread is active but also considered a daemon.
- If the thread was not previously attached, it's now **permanently attached,** blocking shutdown, and cannot later be detached or marked as a daemon

rustls-platform-verifier should instead use the plain old `attach_current_thread`, and recommend to clients to call `attach_current_thread_permanently` for less overhead.

(We ran into this, or think we did, because we used rustls-platform-verifier from a *thread pool* that also attached using plain old `attach_current_thread` elsewhere, and our code was relying on detaching freeing local references. We will probably change that to be explicit anyway, but still, rustls-platform-verifier shouldn't be making the decision.)

PR #185, opened 2025-08-05
Do not permanently attach JVM threads

complexspaces:
We are a library and could be used in any kind of threading context, so we should leave the thread attachment behavior (a performance optimization) choice up to the caller. I did, however, also reduce the number of attachments we call down to 1 as a performance optimization since the existing patterns we have allow it to be safe.

Closes https://github.com/rustls/rustls-platform-verifier/issues/184


complexspaces:
Oh how auspicious, today's the day the mock certificates were set to expire and that's causing new Android failures on an Android-specific changeset 😅:
image

banasiak (approved):
Hello friends. @complexspaces asked me to take a look at this PR. I agree that this is a good idea for the library as it will allow the thread to be released when it goes out of scope. 👍

PR #190, opened 2025-08-13
Update all test certificates and add maintainence documentation

complexspaces:
This PR does several things, with the goal of both unbreaking CI today and making future maintenance easier. These include:
- Reducing the number of chains we need to maintain in the test data
- Simplifying tests with less constants
- Adding documentation for the test certificate update process
- Switching away from 1Password-related certificate chains to ones for AWS themselves. I did this for two reasons:
    - Trying to special-case 1Password less in the repository since its now under the `rustls` org instead.
    - Earlier this year I learned that we (1Password) have not yet enabled TLS 1.3 support for our production servers, so the certificate update script wasn't able to communicate with `my.1password.com` because it only supports TLS 1.3 by default. I could have just enabled the `tls12` feature but it seemed cleaner to just part with it and use a web server that has its TLS configurations updated more often.
    - Updating the real world and mock certificates, as both have since expired this summer.

Commit-by-commit review is recommended.


complexspaces:
Thanks for the fast approval! Please bear with me as I get the various platform-specific compilation failures fixed 😅.

Issue #188, opened 2025-08-06
Unsupported on macOS 10.12


madsmtm:
`rustls-platform-verifier` enables the `OSX_10_14` Cargo feature of `security-framework`, which causes its use to abort on macOS 10.12 ([the minimum supported version by `rustc`](https://doc.rust-lang.org/rustc/platform-support/apple-darwin.html#os-version)) with e.g.:
```
dyld: Symbol not found: _SecTrustEvaluateWithError
  Referenced from: /Users/madsmtm/.cargo/bin/rustup
  Expected in: /System/Library/Frameworks/Security.framework/Versions/A/Security
```

Because the `SecTrustEvaluateWithError` symbol is assumed to be available.

Disabling the feature makes a few error values worse (since `security-framework` now falls back to calling `SecTrustEvaluate`), though it's unclear to me if there's anything else affected by it? Would you accept a PR that does not enable this feature flag?

Git blame reveals this to be present since [the original commit](f9ec702b7353b4a483eb26543f6aa379f62f5a71), CC @complexspaces.

See also https://github.com/kornelski/rust-security-framework/issues/223.

djc:
Also, somewhat relevant that the macOS on x64 target that supports 10.12 is getting demoted to tier 2 (with host tools) soon:

https://github.com/rust-lang/blog.rust-lang.org/pull/1666#discussion_r2236829274

ctz:
I think it is very difficult for us to support platforms that we cannot test in CI, and doing that is especially important if contributors do not and cannot use that combination as their daily-driver.

One thing we could consider here is actually _withdrawing_ platform verifier support for `x86_64-apple-darwin`, and adding it to the list of platforms that use `rustls-native-certs` and the webpki verifier. That has a much smaller platform footprint (so we still have the CI problem, but hopefully on a smaller scale).

PR #194, opened 2025-09-21
Allow macOS 10.12 and 10.13 to work

madsmtm:
Since https://github.com/kornelski/rust-security-framework/pull/243, `security-framework` now dynamically looks up the newer symbol `SecTrustEvaluateWithError`, and only if it's not available does it fall back to `SecTrustEvaluate`. This allows `rustls-platform-verifier` to support macOS 10.12 while still having good error messages on newer versions.

Fixes https://github.com/rustls/rustls-platform-verifier/issues/188.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -315,9 +315,9 @@ dependencies = [
 
 [[package]]
 name = "security-framework"
-version = "3.2.0"
+version = "3.5.0"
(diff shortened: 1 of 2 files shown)


djc (approved):

complexspaces (approved):
This seems like the best path forward, all things considered.

Thank you for putting in the upstream work to make this a clean implementation!

PR #196, opened 2025-10-01
Make cdylib crate type optional

filmor:
Related to https://github.com/rusterlium/rustler/issues/707.

The cdylib is built unconditionally, which leads to it always being included in deployment processes that parse cargo's messages. This change should make the cdylib optional.

### DIFF
--- a/android/rustls-platform-verifier/build.gradle
+++ b/android/rustls-platform-verifier/build.gradle
@@ -50,7 +50,7 @@ android {
 
     task buildTestLib(type: Exec) {
         workingDir "../../"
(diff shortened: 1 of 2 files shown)


djc:
Ah, it looks like this is not going to work?

```rust
error: `crate_type` within an `#![cfg_attr]` attribute is forbidden
 --> rustls-platform-verifier/src/lib.rs:5:33
  |
5 | #![cfg_attr(feature = "cdylib", crate_type = "cdylib")]
  |                                 ^^^^^^^^^^^^^^^^^^^^^

error: could not compile `rustls-platform-verifier` (lib) due to 1 previous error
```

filmor:
Meh, apparently this hack is not possible anymore. What do you actually use the library in? Maybe the respective test-run can be adjusted to explicitly compile a `cdylib` instead.

filmor:
@djc I have adjusted the only user of the cdylib I could find, can you approve the workflow runs?

PR #204, opened 2026-01-02
provide with_platform_verifier_and_extra_roots extension function

### LINKED ISSUE #202: with_extra_roots extension
Currently to use extra roots I need to use the following setup:

```rust
builder.dangerous()
    .with_custom_certificate_verifier(Arc::new(Verifier::new_with_extra_roots(extra_roots)?))
```

There exists the [ext trait](https://github.com/rustls/rustls-platform-verifier/blob/1099f161bfc5e3ac7f90aad88b1bf788e72906cb/rustls-platform-verifier/src/lib.rs#L58-L60) which could be extended to support:

```rust
builder.with_platform_verifier_and_extra_roots(extra_roots)?
```

This avoids needing to justify the use of dangerous when setting up TLS. This mirrors the non-dangerous [`with_root_certificates`](https://docs.rs/rustls/latest/rustls/struct.ConfigBuilder.html#method.with_root_certificates) method.

conradludgate:
Fixes #202

### DIFF
--- a/rustls-platform-verifier/src/lib.rs
+++ b/rustls-platform-verifier/src/lib.rs
@@ -58,6 +58,26 @@ pub trait BuilderVerifierExt {
     fn with_platform_verifier(
         self,
     ) -> Result<ConfigBuilder<ClientConfig, WantsClientCert>, rustls::Error>;
+
+    /// Configures the `ClientConfig` with the platform verifier
+    /// and the extra root certificates to trust.
+    ///
+    /// ```rust
+    /// use rustls::ClientConfig;
+    /// use rustls_platform_verifier::BuilderVerifierExt;
+    ///
+    /// let roots = vec![/* ... */];
+    ///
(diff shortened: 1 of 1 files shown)


djc:
No extra roots on Android... maybe an extension trait method that yields some kind of builder which only offers to set extra roots on non-Android?

conradludgate:
The android/ios tests failing seems to be unrelated. The semver error is correct, but I feel like it is not significant here - no one should be using `dyn BuilderVerifierExt`.

complexspaces (approved):
This looks good to me. I agree the technical semver breakage shouldn't be important. I have a really hard time seeing anyone using the extension traits in their own type declarations since there's only one implementation of each anyway.

complexspaces:
I assume if we merge this the `cargo-semver-checks` job will start to ignore this past "breakage" on future PRs and go back to being green?

PR #210, opened 2026-01-03
fix confusing target doc cfg

### LINKED ISSUE #201: Incorrect documentation for `Verifier`
Spotted in https://github.com/seanmonstar/reqwest/pull/2891#discussion_r2644116022, the documentation appears to only express support for the platform on which the build occurs. For example, https://docs.rs/rustls-platform-verifier/0.6.2/rustls_platform_verifier/struct.Verifier.html#method.new_with_extra_roots:

Image

conradludgate:
fixes #201

Screenshot 2026-01-03 at 15 57 09

The Debug impl is still incorrect, but this is an improvement nonetheless.

### DIFF
--- a/rustls-platform-verifier/src/verification/android.rs
+++ b/rustls-platform-verifier/src/verification/android.rs
@@ -64,6 +64,7 @@ impl Drop for Verifier {
 impl Verifier {
     /// Creates a new instance of a TLS certificate verifier that utilizes the
     /// Android certificate facilities.
+    #[cfg_attr(docsrs, doc(cfg(all())))]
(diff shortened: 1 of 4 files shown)


complexspaces (approved):
This seems like a much better spot to me even if all the cross-platform weirdness isn't gone yet.

IIUC this PR correctly, we are now overriding rustdoc's automatic availability generation and replacing it with manual declarations that `new` and `impl ServerCertVerifier` are available everywhere and that `new_with_extra_roots` is available everywhere but Android all without caring about the underlying OS the docs are built on.

Ralith:
Static indirection through a facade seems like the way to go, yeah.

PR #215, opened 2026-01-17
try with extra roots when root is untrusted

### LINKED ISSUE #214: self-signed end entity cert is not trusted when added as extra root.
https://github.com/rustls/rustls-platform-verifier/blob/93f729c6309680486d33f982e3ba3c42b976f774/rustls-platform-verifier/src/verification/windows.rs#L659

Under windows, extra roots are added only when TrustStatus.dwErrorStatus is CERT_TRUST_IS_PARTIAL_CHAIN. But the error is CERT_TRUST_IS_UNTRUSTED_ROOT when the cert is self-signed and the issuer is itself.

cstkingkey:
Fixes #214


complexspaces:
I've opened #216 to complement this PR and ensure the changed code doesn't regress at minimum.

nab138:
+1, any updates on this?

PR #216, opened 2026-01-29
Add initial new_with_extra_roots testing

complexspaces:
This PR begins adding tests for the `Verifier::new_with_extra_roots` extension currently available on all platforms except Android.

It adds two "simple" cases to cover the known-so-far/common cases we want and expect manually added, non-public roots to behave as trustworthy. I had to modify the `ca.go` generation script to support producing selfsigned certificates that operate on their own without an actual chain.

The second test currently doesn't run on Windows but can be enabled at the same time #215 lands. I validated this locally by manually cherry-picking over the commit:
```
cargo test with_extra_roots
    Finished `test` profile [unoptimized + debuginfo] target(s) in 0.02s
     Running unittests src\lib.rs (target\debug\deps\rustls_platform_verifier-2bee31e54d3b8ab4.exe)

running 2 tests
test tests::verification_mock::test_selfsigned_cert_with_extra_roots ... ok
test tests::verification_mock::test_chain_signed_with_extra_roots ... ok

test result: ok. 2 passed; 0 failed; 0 ignored; 0 measured; 26 filtered out; finished in 0.02s
```


cstkingkey:
As said in https://github.com/rustls/webpki/issues/205#issuecomment-1955469221, self-signed certificates differ in ca parameter. Both cases shall be tested.

[openssl doc](https://docs.openssl.org/master/man5/x509v3_config/#standard-extensions)
> A CA certificate must include the basicConstraints name with the CA parameter set to TRUE. An end-user certificate must either have CA:FALSE or omit the extension entirely.

cstkingkey:
@complexspaces  ca:true is passing on macOS too.

PR #235, opened 2026-06-16
Optionally use rustls-upki for the `other` verifier

ctz:
~~Hm, this is not ideal, as the rustls-upki -> upki -> rustls path turns on unwanted crate features. That means we'd need to infect the upki crate with features also, and make those control the ability to do fetching etc?~~

~~Who designed this stuff 🤣~~

Draft status: needs test in CI

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -29,16 +29,75 @@ dependencies = [
 ]
 
 [[package]]
+name = "anyhow"
+version = "1.0.102"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "7f202df86484c868dbad7eaa557ef785d5c66295e41b460ef922eca0723b842c"
+
+[[package]]
+name = "autocfg"
+version = "1.5.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f2032f911046de80f0a198e0901378627c33f59ea0ac00e363d481118bd70a53"
+
+[[package]]
+name = "aws-lc-rs"
+version = "1.17.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "5ec2f1fc3ec205783a5da9a7e6c1509cc69dedf09a1949e412c1e18469326d00"
+dependencies = [
+ "aws-lc-sys",
+ "untrusted 0.7.1",
+ "zeroize",
+]
+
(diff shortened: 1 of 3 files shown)


complexspaces:
Just from that lockfile diff its quite a lot of new code yeah 😅   

Out of curiosity, if the upki daemon is installed on a system does that also mean the C dylib will be present somewhere on disk too? I wonder if it would be worth considering reusing that code if its already on an end-user's system.

djc:
FYI, this should help:

- https://github.com/rustls/upki/pull/159

> Out of curiosity, if the upki daemon is installed on a system does that also mean the C dylib will be present somewhere on disk too? I wonder if it would be worth considering reusing that code if its already on an end-user's system.

There is no daemon, there would typically be a binary and a systemd unit. (We can also come up with API that can do this in a separate thread or Tokio task -- I think you asked about this a while ago?)

complexspaces:
> There is no daemon, there would typically be a binary and a systemd unit.

Right, bad wording on my side. I was just trying to refer to some binary contents an installer would have put on the system already. IE if there's already a binary would there also be a library with it too for other consumers like OpenSSL?

djc:
> > There is no daemon, there would typically be a binary and a systemd unit.
> 
> Right, bad wording on my side. I was just trying to refer to some binary contents an installer would have put on the system already. IE if there's already a binary would there also be a library with it too for other consumers like OpenSSL?

Basically yes, see https://launchpad.net/~ubuntu-foundations-team/+archive/ubuntu/upki-preview/+packages. But I think it would still be preferable to use Rust interfaces here?

complexspaces:
I have mixed feelings on it so far 😄. On every other platform that provides certificate verification facilities, we use the shared libraries provided by the "platform" instead of re-implementing or shipping logic that just uses raw data the OS holds somewhere for itself. If you squint enough on Linux, the upki infrastructure installed by the distro would count as those verification facilities.

djc:
> I have mixed feelings on it so far 😄. On every other platform that provides certificate verification facilities, we use the shared libraries provided by the "platform" instead of re-implementing or shipping logic that just uses raw data the OS holds somewhere for itself. If you squint enough on Linux, the upki infrastructure installed by the distro would count as those verification facilities.

What are you concretely worried about?

PR #238, opened 2026-07-08
windows: add Verifier::offline() constructor

### LINKED ISSUE #237: Windows: revocation check blocks up to 10s with no effect on verification outcome
## Summary

On Windows, `CertGetCertificateChain` is called with `CERT_CHAIN_REVOCATION_CHECK_END_CERT`, which triggers a network call to OCSP/CRL endpoints with a 10-second timeout. However, in `verify_chain_policy`, `CERT_CHAIN_POLICY_IGNORE_ALL_REV_UNKNOWN_FLAGS` is set, meaning revocation failures are silently ignored in the final policy decision.

This results in a blocking network call of up to 10 seconds that has no effect on whether the certificate is ultimately accepted or rejected.

## Impact

In network-restricted environments (e.g., hosts with limited outbound access where OCSP/CRL endpoints are unreachable), the TLS handshake blocks for the full 10-second `dwUrlRetrievalTimeout` on every first connection attempt. This is particularly problematic f
or:

- Short-lived instances where the CRL cache is always cold
- Hosts behind firewalls that allow :443 to the application server but not to arbitrary OCSP responders
- Applications with request timeouts shorter than 10 seconds (the revocation timeout exceeds the application timeout, causing connection failures)

One can hit this e.g. after migrating from `webpki-roots` (no network calls) to `rustls-platform-verifier` (via reqwest 0.13).

## Relevant code

In `src/verification/windows.rs`:

```rust
// Chain building — triggers network call
const FLAGS: u32 = CERT_CHAIN_REVOCATION_CHECK_END_CERT
    | CERT_CHAIN_REVOCATION_ACCUMULATIVE_TIMEOUT
    | CERT_CHAIN_CACHE_END_CERT;

parameters.dwUrlRetrievalTimeout = 10 * 1000; // 10 seconds
```

```rust
// Policy verification — ignores revocation failures
params.dwFlags = CERT_CHAIN_POLICY_IGNORE_ALL_REV_UNKNOWN_FLAGS;
```

## Comparison with other implementations

- **Go `crypto/x509`** (`root_windows.go`): Calls `CertGetCertificateChain` without any `CERT_CHAIN_REVOCATION_CHECK_*` flags - no revocation network calls.
- **The test configuration in this crate**: Uses `CERT_CHAIN_CACHE_ONLY_URL_RETRIEVAL` to prevent network calls, which confirms the production code intentionally makes them.

## Suggestion

Would you consider one of the following?

1. **Remove `CERT_CHAIN_REVOCATION_CHECK_END_CERT`** from the flags, since the result is discarded by the policy step anyway. This eliminates the network call with no change in security posture.

2. **Make revocation checking configurable** via a builder option on `Verifier` (e.g., `with_revocation_check(bool)` or `with_revocation_check(RevocationPolicy)`) so consumers can opt out in environments where the network call is problematic.

3. **Use `CERT_CHAIN_REVOCATION_CHECK_CACHE_ONLY`** by default (check cached CRLs without making network calls), with an opt-in for full network-based revocation.

## Environment

- Windows Server 2025
- rustls-platform-verifier 0.7.x
- reqwest 0.13

djc:
Does a bunch of cleanup/refactoring first, and tightens the timeout from 10 to 3s.

Fixes #237.

### DIFF
--- a/rustls-platform-verifier/src/verification/windows.rs
+++ b/rustls-platform-verifier/src/verification/windows.rs
@@ -88,7 +88,7 @@ impl Verifier {
 
     /// Creates a new instance of a TLS certificate verifier that utilizes the
     /// Windows certificate facilities and augmented by the provided extra root certificates.
-    #[cfg_attr(docsrs, doc(cfg(not(target_os = "android"))))]
+    #[cfg_attr(docsrs, doc(cfg(all())))]
     pub fn new_with_extra_roots(
         roots: impl IntoIterator<Item = pki_types::CertificateDer<'static>>,
         crypto_provider: Arc<CryptoProvider>,
--- a/rustls-platform-verifier/src/verification/windows.rs
+++ b/rustls-platform-verifier/src/verification/windows.rs
@@ -129,7 +129,7 @@ impl Verifier {
         #[cfg(any(test, feature = "ffi-testing", feature = "dbg"))]
         let mut store = match self.test_only_root_ca_override.as_ref() {
             Some(test_only_root_ca_override) => {
-                CertificateStore::new_with_fake_root(test_only_root_ca_override.clone())?
+                CertificateStore::for_testing(test_only_root_ca_override.clone())?
             }
             None => CertificateStore::new()?,
         };
@@ -304,7 +304,7 @@ impl CertEngine {
     }
 
     #[cfg(any(test, feature = "ffi-testing", feature = "dbg"))]
-    fn new_with_fake_root(root: pki_types::CertificateDer<'static>) -> Result<Self, TlsError> {
+    fn for_testing(root: pki_types::CertificateDer<'static>) -> Result<Self, TlsError> {
         // We use these flags for the following reasons:
         //
         // - CERT_CHAIN_CACHE_ONLY_URL_RETRIEVAL is used in an attempt to stop Windows from using the internet to
@@ -383,11 +383,11 @@ struct CertificateStore {
 
 impl CertificateStore {
     #[cfg(any(test, feature = "ffi-testing", feature = "dbg"))]
-    fn new_with_fake_root(root: pki_types::CertificateDer<'static>) -> Result<Self, TlsError> {
+    fn for_testing(root: pki_types::CertificateDer<'static>) -> Result<Self, TlsError> {
         let mut inner = Self::new()?;
         inner.add_cert(&root)?;
 
-        let engine = CertEngine::new_with_fake_root(root)?;
+        let engine = CertEngine::for_testing(root)?;
         inner.engine = Some(engine);
 
         Ok(inner)
--- a/rustls-platform-verifier/src/verification/windows.rs
+++ b/rustls-platform-verifier/src/verification/windows.rs
@@ -300,37 +300,11 @@ impl CertEngine {
     fn new_with_extra_roots(
         roots: impl IntoIterator<Item = pki_types::CertificateDer<'static>>,
     ) -> Result<Self, TlsError> {
-        let mut exclusive_store = CertificateStore::new()?;
-        for root in roots {
-            exclusive_store.add_cert(&root)?;
-        }
-
-        let mut config = CERT_CHAIN_ENGINE_CONFIG::zeroed_with_size();
-        config.hExclusiveRoot = exclusive_store.inner.as_ptr();
-
-        let mut engine = EnginePtr::NULL;
-
-        // XXX: Due to the redefinition of `CERT_CHAIN_ENGINE_CONFIG`, we need to do pointer casts
-        // in order to pass our expanded structure into `CertCreateCertificateChainEngine`.
-        // See also `CERT_CHAIN_PARA` casting below.
-        let config = NonNull::from(&config).cast().as_ptr();
-        // SAFETY: `engine` is valid to be written to and the config is valid to be read.
-        let res = unsafe { CertCreateCertificateChainEngine(config, &mut engine) };
-
-        #[allow(clippy::as_conversions)]
-        let engine = call_with_last_error(|| match NonNull::new(engine as *mut c_void) {
-            Some(c) if res == TRUE => Some(c),
-            _ => None,
-        })?;
-        Ok(Self { inner: engine })
+        Self::new(roots, 0)
     }
 
     #[cfg(any(test, feature = "ffi-testing", feature = "dbg"))]
     fn new_with_fake_root(root: pki_types::CertificateDer<'static>) -> Result<Self, TlsError> {
-        let mut root_store = CertificateStore::new()?;
-        root_store.add_cert(&root)?;
-
-        let mut config = CERT_CHAIN_ENGINE_CONFIG::zeroed_with_size();
         // We use these flags for the following reasons:
         //
         // - CERT_CHAIN_CACHE_ONLY_URL_RETRIEVAL is used in an attempt to stop Windows from using the internet to
@@ -340,11 +314,29 @@ impl CertEngine {
         // data inside of a test and avoid any extra parsing, etc, it might need to do pulling directly from the store each time.
         //
         // Ref: https://docs.microsoft.com/en-us/windows/win32/api/wincrypt/ns-wincrypt-cert_chain_engine_config
-        config.dwFlags = CERT_CHAIN_CACHE_ONLY_URL_RETRIEVAL | CERT_CHAIN_ENABLE_CACHE_AUTO_UPDATE;
-        config.hExclusiveRoot = root_store.inner.as_ptr();
+        Self::new(
+            [root],
+            CERT_CHAIN_CACHE_ONLY_URL_RETRIEVAL | CERT_CHAIN_ENABLE_CACHE_AUTO_UPDATE,
+        )
+    }
 
+    fn new(
+        roots: impl IntoIterator<Item = pki_types::CertificateDer<'static>>,
+        flags: u32,
+    ) -> Result<Self, TlsError> {
+        let mut store = CertificateStore::new()?;
+        for root in roots {
+            store.add_cert(&root)?;
+        }
+
+        let mut config = CERT_CHAIN_ENGINE_CONFIG::zeroed_with_size();
+        config.hExclusiveRoot = store.inner.as_ptr();
+        config.dwFlags = flags;
         let mut engine = EnginePtr::NULL;
-        // Same workaround with as above when creating the engine.
+
+        // XXX: Due to the redefinition of `CERT_CHAIN_ENGINE_CONFIG`, we need to do pointer casts
+        // in order to pass our expanded structure into `CertCreateCertificateChainEngine`.
+        // See also `CERT_CHAIN_PARA` casting below.
         let config = NonNull::from(&config).cast().as_ptr();
         // SAFETY: `engine` is valid to be written to and the config is valid to be read.
         let res = unsafe { CertCreateCertificateChainEngine(config, &mut engine) };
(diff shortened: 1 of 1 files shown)


lpapp:
Thanks for the quick response on this, @djc!

The `Verifier::offline()` constructor gets us most of the way there, but I think it is more restrictive than necessary. `CERT_CHAIN_CACHE_ONLY_URL_RETRIEVAL` disables **all** network fetches - including AIA (Authority Information Access) chasing for intermediate certificates, not just revocation.

For comparison, Go's `crypto/x509` on Windows calls `CertGetCertificateChain` with no `CERT_CHAIN_REVOCATION_CHECK_*` flags and no `CERT_CHAIN_CACHE_ONLY_URL_RETRIEVAL`. This gives you:
- AIA fetching (intermediates downloaded if the server does not send the full chain) 
- No revocation network calls
- Platform trust store (no need to supply roots manually)

Would you consider a mode that simply omits `CERT_CHAIN_REVOCATION_CHECK_END_CERT` from the flags instead of using `CERT_CHAIN_CACHE_ONLY_URL_RETRIEVAL`? That would be the more surgical fix - it eliminates the revocation timeout without breaking AIA or requiring the caller to supply their own root certificates.

As it stands, `offline()` requires the caller to pass in roots explicitly, which removes one of the main benefits of using `rustls-platform-verifier` over `webpki-roots` in the first place (automatic use of the system trust store).

djc:
Do you want to send a PR against my branch that better fits your requirements? We should potentially rethink the `offline()` name to better match the semantics of what you're trying to achieve, too.

lpapp:
I am fine with this PR. We can revisit with more granular control in the future if needed. Appreciate the quick turnaround. Do you have a sense of when this might be merged and released?

djc:
> I am fine with this PR. We can revisit with more granular control in the future if needed. Appreciate the quick turnaround. Do you have a sense of when this might be merged and released?

Reviews in this repo have been a little slow recently. @complexspaces will you have time to review this in the next, say, two weeks or so?

complexspaces:
Yeah I will make time this week 👍. Thanks for putting some refactoring work into this as well.

I wanted to take a second look at the flags being used here since I was confused by this comment in the linked issue:
> Remove CERT_CHAIN_REVOCATION_CHECK_END_CERT from the flags, since the result is discarded by the policy step anyway. This eliminates the network call with no change in security posture.

lpapp:
Hi @complexspaces, just a friendly nudge - since today is the last working day of the week, would you have a chance to take a look at this? A release with this fix would unblock some downstream work. No pressure of course, and thanks again for taking the time to review!

rustls-platform-verifier/src/verification/windows.rs
```diff
@@ -55,185 +57,249 @@ use windows_sys::Win32::{
...
-}
 
-// SAFETY: see method implementation
-unsafe impl ZeroedWithSize for CERT_CHAIN_POLICY_PARA {
-    fn zeroed_with_size() -> Self {
-        // SAFETY: This structure only contains integers and pointers.
-        let mut new: Self = unsafe { mem::zeroed() };
-        new.cbSize = Self::SIZE;
-        new
+    /// Creates a Windows TLS certificate verifier that avoids using online revocation checking.
```
briansmith, line 103: I would expect an  `offline` constructor to avoid any/all network fetches, not just revocation checks. If the goal is just to avoid OCSP/CRL fetching, then a different name should be used. Regardless, I think it is also useful to state more explicitly what this does for AIA fetching.

However, IDK if the Windows certificate verification API is practically usable in a purely offline mode, in general, since Windows usually fetches the trust anchors themselves dynamically, IIRC.

lpapp: > However, IDK if the Windows certificate verification API is practically usable in a purely offline mode, in general, since Windows usually fetches the trust anchors themselves dynamically, IIRC.

For environments where trust anchors are provisioned into the local cert store, offline mode works fine.

briansmith: > For environments where trust anchors are provisioned into the local cert store, offline mode works fine.

That's a minority of systems, though, isn't it? My understanding is that on most systems, the root store starts almost empty OOB.

lpapp: I do not have data on what percentage of Windows systems rely on dynamic root fetching vs. pre-provisioned stores. This is an argument for making the behavior configurable rather than a single `offline()` toggle. If no revocation (omit `CERT_CHAIN_REVOCATION_CHECK_END_CERT`, leave everything else alone) - eliminates revocation network calls while preserving AIA and dynamic root fetching. This matches Go's `crypto/x509` behavior on Windows.

That said, if splitting these into separate modes is a significant effort from the author, I think the current PR is still a clear improvement over the status quo - it enables flows where both revocation and AIA are unnecessary, which is a real use case today.

complexspaces:
@lpapp Yup, I am reviewing this today. I'm currently doing some local testing as well on Windows.

rustls-platform-verifier/src/verification/windows.rs
complexspaces: > @briansmith: However, IDK if the Windows certificate verification API is practically usable in a purely offline mode, in general, since Windows usually fetches the trust anchors themselves dynamically, IIRC.

It is hard for me to tell exactly how much AIA is needed (especially with MS's own certificate fetching logic in Windows Update) but from what I can see on my own Windows systems I would lean towards scoping this change to just disallow networking for revocation specifically.

To provide an example: On my Windows 10 system, I only have LetsEncrypt's X1 root by default. X2 is not in any machine/user stores. In that state, I tested a few different flag configurations of `CertGetCertificateChain` (which is the function that actually fetches certificates that aren't present) when no intermediates from `rustls` get passed to Windows.

I used `reqwest` to fetch `https://letsencrypt.org` which uses a certificate from their X2 root. I also added `CERT_CHAIN_CACHE_ONLY_URL_RETRIEVAL` to every call of `CertGetCertificateChain` as a baseline then changed the below to see the different behaviors:

| CERT_CHAIN_DISABLE_AIA | CERT_CHAIN_DISABLE_AUTH_ROOT_AUTO_UPDATE | Result |
|-------------------------------|---------------------------------------------------------|---------|
| No                    | No   | Root is added to `Intermediate Certification Authorities` store |
| No                    | Yes    | Root is added to `Intermediate Certification Authorities` store by AIA |
| Yes                    |  No |   Root is added to `Third-Party Root Certification Authorities` store by Windows Update(?)
| Yes                   | Yes |  Fails to verify certificate due to unknown issuer

Similar applies to the `YE` intermediate root but I didn't spend the time cleaning that one up too between runs, just X2.

Given that there is a complete failure in a very common case, I think that its safe to say that an entirely offline Windows verifier is not a useful tool to expose from this crate. It won't be able to load a vast majority of sites on the internet.

Maybe it would be best to rename this to `cached_revocation` or `offline_revocation`?

> @lpapp: For environments where trust anchors are provisioned into the local cert store, offline mode works fine.

I would be surprised if your provisioning process is set up in such a way that machines are always given a fresh copy of the Windows CAB containing all known trusted roots on a regular basis. 

I've also ran into a case at work where someone wasn't properly configuring their VDI infrastructure with root certificates so everything 1Password tried to do online failed. I air on the side of assuming the machine won't arrive in an offline-usable state for third-party desktop apps.

rustls-platform-verifier/src/verification/windows.rs
```diff
@@ -466,7 +466,7 @@ impl CertificateStore {
             | CERT_CHAIN_CACHE_END_CERT;
 
         // Lowering URL retrieval timeout from default 15s to 10s to account for higher internet speeds
```
complexspaces, line 468: Nit: We can either update this comment or just delete it outright now.

rustls-platform-verifier/src/verification/windows.rs
```diff
@@ -326,9 +326,9 @@ impl CertEngine {
     }
 
     #[cfg(any(test, feature = "ffi-testing", feature = "dbg"))]
-    fn new_with_fake_root(root: &[u8]) -> Result<Self, TlsError> {
+    fn new_with_fake_root(root: pki_types::CertificateDer<'static>) -> Result<Self, TlsError> {
```
complexspaces, line 329: Is there a reason this doesn't take a reference to `CertificateDer` if it only needs that? It would avoid the `.clone()` noise.

rustls-platform-verifier/src/verification/windows.rs
```diff
@@ -391,11 +391,11 @@ struct CertificateStore {
 
 impl CertificateStore {
     #[cfg(any(test, feature = "ffi-testing", feature = "dbg"))]
-    fn new_with_fake_root(root: &[u8]) -> Result<Self, TlsError> {
+    fn new_with_fake_root(root: pki_types::CertificateDer<'static>) -> Result<Self, TlsError> {
```
complexspaces, line 394: Ditto here.

rustls-platform-verifier/src/verification/windows.rs
```diff
@@ -102,6 +100,20 @@ impl Verifier {
         })
     }
 
+    /// Creates a Windows TLS certificate verifier that avoids using online revocation checking.
+    #[cfg_attr(docsrs, doc(cfg(all())))]
+    pub fn offline(
```
complexspaces, line 105: What do you think about making this switchable with a flag setter instead of a whole new constructor? I think its makes the API more confusing exposing the `roots` parameter outside of the `new_with_extra_roots` constructor, in addition to a whole new constructor option.

The downside of doing it that way would be that the `set_offline_mode(&mut self, offline: bool)` function would need to call `CertCreateCertificateChainEngine` again and replace the existing one. But everywhere else we could more lazily obtain the flag value and not need any more Win32 calls to accommodate.

rustls-platform-verifier/src/verification/windows.rs
```diff
@@ -102,6 +100,20 @@ impl Verifier {
...
+        roots: impl IntoIterator<Item = pki_types::CertificateDer<'static>>,
+        crypto_provider: Arc<CryptoProvider>,
+    ) -> Result<Self, TlsError> {
+        Ok(Self {
+            #[cfg(any(test, feature = "ffi-testing", feature = "dbg"))]
+            test_only_root_ca_override: None,
+            crypto_provider,
+            extra_roots: Some(CertEngine::new(roots, CERT_CHAIN_CACHE_ONLY_URL_RETRIEVAL)?),
```
complexspaces, line 113: Its probably best to keep the discussion in https://github.com/rustls/rustls-platform-verifier/pull/238#discussion_r3604813184, but per my findings there setting this isn't actually enough to stop Windows from doing _all_ networking during the verification calls FYI.

complexspaces (changes requested):

rustls-platform-verifier/src/verification/windows.rs
lpapp: > I would be surprised if your provisioning process is set up in such a way that machines are always given a fresh copy of the Windows CAB containing all known trusted roots on a regular basis.

This setup does happen.

briansmith: > This setup does happen.

I think it is useful to have a pure no-networking configuration. But, I think that the goal is to fix #237, which requires a different solution and different naming.

lpapp: To clarify my use case from #237 - I need to disable revocation checks specifically. Whether AIA remains enabled or disabled is not a concern for my scenario, so either the current `offline()` approach or a revocation-only toggle would resolve the issue for me. The current one week delay is more concerning to me because this is a high-priority blocker for my project, so I am now also evaluating rustls alternatives.

briansmith: @lpapp Being annoying is unlikely to speed up the process.

PR #241, opened 2026-07-29
Fix Windows compatibility with hostnames containing a trailing label

### LINKED ISSUE #240: Windows: an absolute FQDN (trailing dot) fails NotValidForName where webpki accepts it
## Summary

On Windows, a `ServerName` written as an absolute FQDN — with the root
label's trailing dot — fails verification with `NotValidForName`, even
though the certificate is valid for that host and `WebPkiServerVerifier`
accepts the exact same name against the exact same certificate.

The name is handed to `CertVerifyCertificateChainPolicy` verbatim, and
CryptoAPI compares `example.com.` against the certificate's dot-less
SAN.

This makes swapping `WebPkiServerVerifier` for `Verifier` a behaviour
change rather than a trust-store change, for anyone whose server names
come from a source that keeps the root label.

## Reproducer

Nothing exotic — two public hosts, both verifiers, with and without the
trailing dot.

```toml
[dependencies]
rustls = { version = "0.23", default-features = false, features = ["ring", "std", "logging"] }
rustls-platform-verifier = "0.7"
rustls-pki-types = "1"
webpki-roots = "1"
```

```rust
use std::io::{Read, Write};
use std::sync::Arc;

fn probe(host: &str, name: &'static str, system: bool) {
    let provider = Arc::new(rustls::crypto::ring::default_provider());
    let verifier: Arc<dyn rustls::client::danger::ServerCertVerifier> = if system {
        Arc::new(rustls_platform_verifier::Verifier::new(provider.clone()).unwrap())
    } else {
        let roots = Arc::new(rustls::RootCertStore {
            roots: webpki_roots::TLS_SERVER_ROOTS.to_vec(),
        });
        rustls::client::WebPkiServerVerifier::builder_with_provider(roots, provider.clone())
            .build()
            .unwrap()
    };
    let config = rustls::ClientConfig::builder_with_provider(provider)
        .with_safe_default_protocol_versions()
        .unwrap()
        .dangerous()
        .with_custom_certificate_verifier(verifier)
        .with_no_client_auth();
    let sn = rustls_pki_types::ServerName::try_from(name).unwrap();
    let mut conn = rustls::ClientConnection::new(Arc::new(config), sn).unwrap();
    let mut sock = std::net::TcpStream::connect((host, 443)).unwrap();
    let mut tls = rustls::Stream::new(&mut conn, &mut sock);
    let out = tls
        .write_all(format!("GET / HTTP/1.0\r\nHost: {host}\r\n\r\n").as_bytes())
        .and_then(|_| {
            let mut b = Vec::new();
            let _ = tls.read_to_end(&mut b);
            Ok(())
        });
    let label = if system { "system  " } else { "embedded" };
    match out {
        Ok(()) => println!("{label} + {name:24} -> OK"),
        Err(e) => println!("{label} + {name:24} -> {e}"),
    }
}

fn main() {
    for host in ["example.com", "www.rust-lang.org"] {
        for system in [false, true] {
            probe(host, Box::leak(host.to_string().into_boxed_str()), system);
            probe(host, Box::leak(format!("{host}.").into_boxed_str()), system);
        }
    }
}
```

Output on Windows 11 (26200), `rustls-platform-verifier` 0.7.0:

```
embedded + example.com              -> OK
embedded + example.com.             -> OK
system   + example.com              -> OK
system   + example.com.             -> invalid peer certificate: NotValidForName
embedded + www.rust-lang.org        -> OK
embedded + www.rust-lang.org.       -> OK
system   + www.rust-lang.org        -> OK
system   + www.rust-lang.org.       -> invalid peer certificate: NotValidForName
```

## Cause

`src/verification/windows.rs`:

```rust
let name = server_name.to_str();
// …
self.verify_certificate(end_entity.as_ref(), &intermediate_certs, name.as_bytes(), …)
```

which becomes `HTTPSPolicyCallbackData::pwszServerName` for
`CertVerifyCertificateChainPolicy(CERT_CHAIN_POLICY_SSL, …)`. That API
does a literal comparison against the SAN entries, so the root label has
to be gone by the time it gets there. The returned
`CERT_E_INVALID_NAME` / `CERT_E_CN_NO_MATCH` is then mapped to
`CertificateError::NotValidForName`.

webpki drops the root label as part of name matching, which is why the
embedded path is unaffected.

Worth noting the name is valid at every layer above: `ServerName::try_from`
accepts it, and SChannel does too — .NET's `SslStream.AuthenticateAsClient`
against the same host with the same trailing-dot target host succeeds on
this machine, so it evidently normalises before reaching the same policy
check.

## Suggested fix

Strip a single trailing `.` from the name before handing it to the
platform API. `src/verification/apple.rs` passes `server_name.to_str()`
straight to `SecPolicy::create_ssl` the same way — I have no macOS box to
check whether Security.framework tolerates it, so that half is a guess,
but the shape is identical.

## How I ran into it

[iroh](https://github.com/n0-computer/iroh) writes its default relay
hostnames as absolute FQDNs (`aps1-1.relay.n0.iroh.link.`), and offers a
`platform-verifier` feature so users behind a TLS-inspecting corporate
proxy can trust the machine's own roots. Turning that feature on makes
every relay fail `NotValidForName` — the one situation the feature exists
for is the one where it cannot be used. Filing separately over there.

## Environment

- Windows 11 Pro 26200
- `rustls-platform-verifier` 0.7.0, `rustls` 0.23, ring backend

complexspaces:
This PR fixes a small incompatibility in the Windows verifier around hostnames which have a trailing `.` in them that originates from the Win32 API itself.

Other platforms do not appear to have the same limitation but I added the same stripping change everywhere to help keep the Rust layer of this library consistent cross-platform per its goals.

Closes #240

### DIFF
--- a/rustls-platform-verifier/src/tests/verification_mock/mod.rs
+++ b/rustls-platform-verifier/src/tests/verification_mock/mod.rs
@@ -152,6 +152,16 @@ mock_root_test_cases! {
         expected_result: Ok(()),
         other_error: no_error!(),
     },
+    valid_no_stapling_dns_trailing_label [ any(windows, unix) ] => TestCase {
+        // XXX: `pki_types` validates that the label is still valid and that cases such as two `.` characters
+        // are still correctly rejected.
+        reference_id: "example.com.",
+        chain: &[ROOT1_INT1_EXAMPLE_COM_GOOD, ROOT1_INT1],
+        stapled_ocsp: None,
+        verification_time: verification_time(),
+        expected_result: Ok(()),
+        other_error: no_error!(),
+    },
     valid_no_stapling_ipv4 [ any(windows, unix) ] => TestCase {
         reference_id: LOCALHOST_IPV4,
         chain: &[ROOT1_INT1_LOCALHOST_IPV4_GOOD, ROOT1_INT1],
--- a/rustls-platform-verifier/src/verification/android.rs
+++ b/rustls-platform-verifier/src/verification/android.rs
@@ -91,7 +91,7 @@ impl Verifier {
         &self,
         end_entity: &pki_types::CertificateDer<'_>,
         intermediates: &[pki_types::CertificateDer<'_>],
-        server_name: &pki_types::ServerName<'_>,
+        server_name: &pki_types::ServerName,
         ocsp_response: Option<&[u8]>,
         now: pki_types::UnixTime,
     ) -> Result<(), TlsError> {
@@ -172,7 +172,11 @@ impl Verifier {
                 ) -> org.rustls.platformverifier.VerificationResult
             );
 
-            let server_name = cx.env.new_string(server_name.to_str())?;
+            let server_name = server_name.to_str();
+            // Android's verifier doesn't require this but trim trailing `.` labels for consistency across platforms.
+            let server_name = server_name.strip_suffix('.').unwrap_or(&server_name);
+
+            let server_name = cx.env.new_string(server_name)?;
             let auth_type = cx.env.new_string(AUTH_TYPE)?;
 
             let result = cx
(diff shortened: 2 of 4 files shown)


djc (approved):
Makes sense to me.

Issue #221, opened 2026-02-19
Android verifier returns Revoked for Let's Encrypt certificates (CRL-only, no OCSP URL)


ospfranco:
I'm facing an error on Android where a request (done via reqwest) to https://sgx.opacitylabs.com:18081/sgx/certification/v4/qe/identity gives me a `invalid peer certificate: Revoked` error when using `rustls-platform-verifier`. After asking claude to spelunk a bit it came up with the following info

```
Since [Let's Encrypt ended OCSP support in August 2025](https://letsencrypt.org/2025/08/06/ocsp-service-has-reached-end-of-life), all newly issued Let's Encrypt certificates no longer include an OCSP URL in the Authority Information Access extension. These certificates only include a CRL Distribution Point. On Android, rustls-platform-verifier returns invalid peer certificate: Revoked for these certificates, even though the certificate is valid and not present in the CRL.


Steps to Reproduce
Use rustls-platform-verifier on Android to connect to any host with a Let's Encrypt certificate issued after August 2025.

The certificate's AIA extension will look like this (no OCSP entry):


Authority Information Access:
    CA Issuers - URI:http://e7.i.lencr.org/
Instead of the previously expected format:


Authority Information Access:
    OCSP - URI:http://e7.o.lencr.org/
    CA Issuers - URI:http://e7.i.lencr.org/
Expected Behavior
Certificate validation succeeds, falling back to CRL-based revocation checking or skipping revocation if CRL is also unavailable.

Actual Behavior

invalid peer certificate: Revoked
Root Cause
Android's TrustManager performs mandatory OCSP revocation checking for publicly-trusted certificates. When the certificate has no OCSP URL, it appears to fail with a Revoked error rather than falling back to the CRL distribution point or treating revocation as unknown. rustls-platform-verifier surfaces this as CertificateError::Revoked.

This is not specific to any one certificate or domain — it affects all Let's Encrypt certificates going forward, as LE now exclusively publishes revocation information via CRLs.

Environment
rustls-platform-verifier: 0.6.2
Android (all versions tested)
Let's Encrypt certificates issued after August 2025 (intermediates E5, E6, E7, E8, R10, R11, R12)
```

I'm not 100% certain this is correct but we have checked our certificates and everything comes from Let's Encrypt and they are not revoked, so I'm not sure what else could be wrong.

elichai:
[...]

PR #246, opened 2026-08-26
android: advance the alias counter before continuing in isKnownRoot

### LINKED ISSUE #245: Android: `isKnownRoot` never terminates when a system trust anchor has been disabled by the user
## Summary

`CertificateVerifier.isKnownRoot` walks the collision-suffixed aliases of a system trust anchor. Two of the three branches inside the loop `continue` without advancing the loop counter, so the loop re-evaluates the exact same alias forever. The thread that entered TLS verification repeats the same filesystem and keystore lookups indefinitely and never returns.

This is an availability bug, not a security one: no verification decision is weakened, the call simply never completes.

## Location

`android/rustls-platform-verifier/src/main/java/org/rustls/platformverifier/CertificateVerifier.kt`, `isKnownRoot`, lines 431-462 at the current tip of `main` (last commit touching this file: `042330d85e8f14f8a088f6522718fcde5fb6da8d`).

```kotlin
var i = 0
while (true) {
    val alias = "$hash.$i"

    if (!File(loadedSystemCertificateDirectory, alias).exists()) {
        break
    }

    val anchor = loadedSystemKeystore.getCertificate("system:$alias")

    // It's possible for `anchor` to be `null` if the user deleted a trust anchor.
    // Continue iterating as there may be further collisions after the deleted anchor.
    if (anchor == null) {
        continue
        // This should never happen
    } else if (anchor !is X509Certificate) {
        // SAFETY: ...
        Log.e(TAG, "anchor is not a certificate, alias: $alias")
        continue
        // If subject and public key match, it's a system root.
    } else {
        if ((root.subjectX500Principal == anchor.subjectX500Principal) && (root.publicKey == anchor.publicKey)) {
            systemTrustAnchorCache.add(key)
            return true
        }
    }

    i += 1
}
```

`i += 1` is the last statement of the loop body. Both `continue` statements jump over it. When either branch is taken:

- `i` is unchanged, so `alias` is recomputed to the same string,
- the same file still exists, so the loop does not `break`,
- `getCertificate("system:$alias")` returns the same value, so the same branch is taken again.

The comment directly above the first `continue` ("Continue iterating as there may be further collisions after the deleted anchor") shows the intent was to advance to the next alias.

For reference, the Chromium code this is ported from (`net/android/java/src/org/chromium/net/X509Util.java`) puts the increment in the `for` update clause:

```java
for (int i = 0; true; i++) {
    String alias = hash + '.' + i;
    ...
    if (anchor == null) continue;
```

There, `continue` still advances `i`. Kotlin has no C-style `for`, and the port moved the increment into the body, which changed what `continue` does.

## Reproduction

The `anchor == null` branch is reachable exactly as its own comment describes: the alias file is present under the system cacerts directory but `AndroidCAStore` has no entry for it.

That is the state of a device where the user has turned off a preinstalled system CA (Settings > Security > Encryption & credentials > Trusted credentials > System > *disable*). The directory this file scans is `$ANDROID_ROOT/etc/security/cacerts` (resolved via `System.getenv("ANDROID_ROOT")`, normally `/system`), which is read-only, so `<hash>.<i>` stays on disk. The removal is recorded elsewhere: conscrypt's `TrustedCertificateStore` copies the certificate under `$ANDROID_DATA/misc/keychain/cacerts-removed`, and `getCertificate` returns `null` for a system alias once `isDeletedSystemCertificate` matches.

The upstream Chromium comment describes the same mechanism explicitly -- "the certificate remains in the system directory but is also added to another file" -- which is why its loop was written to keep advancing past that alias.

So: on such a device, verify any chain whose root subject hashes to the alias of the disabled anchor. `verifyCertificateChain` calls `isKnownRoot(validChain.last())` (in the `ocspResponse == null` revocation-gating path, API >= 24) and the call never returns.

I found this by code review while vendoring the file rather than from a field report, so I have not captured a trace from a physical device; the reachability argument above is the whole of the evidence.

## Impact

- The affected thread is whichever thread `rustls` is running the handshake on. It never completes, and keeps a core busy re-running the same `File.exists()` / `getCertificate` pair; there is no timeout and no way for the caller to recover.
- The second `continue` (`anchor !is X509Certificate`) has the same defect and additionally writes a log line on every iteration, so it would also flood logcat.

## Suggested fix

Advance the counter before any branch can `continue`:

```diff
                 var i = 0
                 while (true) {
                     val alias = "$hash.$i"
+                    i += 1
 
                     if (!File(loadedSystemCertificateDirectory, alias).exists()) {
                         break
@@
                             return true
                         }
                     }
-
-                    i += 1
                 }
```

`i` is not read after the loop, so incrementing before the `break` check is harmless, and progress no longer depends on how many `continue` paths the body has. `val alias = "$hash.${i++}"` is an equivalent one-line alternative.

Happy to open a PR with either form if that helps.

ChoiSeungMyung:
Fixes #245.

`isKnownRoot` walks the collision-suffixed aliases of a system trust anchor. `i += 1` was the last statement of the loop body, and two of the three branches reach the next iteration through `continue`, which jumps over it. When either branch is taken:

- `i` is unchanged, so `alias` is recomputed to the same string,
- the same file still exists, so the loop does not `break`,
- `getCertificate("system:$alias")` returns the same value, so the same branch is taken again.

The thread running the handshake never returns and keeps a core busy repeating one `File.exists()` and one `getCertificate()` call. The `anchor !is X509Certificate` branch additionally logs on every iteration.

This is an availability bug, not a security one — no verification decision is weakened, the call simply never completes.

### Why the branch is reachable

Exactly as its own comment says. conscrypt's `TrustedCertificateStore.getCertificate` returns `null` for a system alias once `isDeletedSystemCertificate` matches, which happens after the certificate is copied into `$ANDROID_DATA/misc/keychain/cacerts-removed` — the state produced by disabling a preinstalled CA in Settings. The file under `$ANDROID_ROOT/etc/security/cacerts` stays on disk because that partition is read-only, so `File.exists()` still returns true.

The Chromium comment this was ported from spells out the same mechanism: *"the certificate remains in the system directory but is also added to another file. Continue iterating as there may be further collisions after the deleted anchor."*

`isKnownRoot` is reached from `verifyCertificateChain` at the `ocspResponse == null` revocation-gating path (API >= 24), so any chain whose root subject hashes to the disabled anchor's alias hits this.

### The fix

Chromium keeps the increment in the update clause of a C-style `for`, so `continue` still advances there. Kotlin has no such form and the port moved it into the body, changing what `continue` does. Incrementing right after the alias is built restores that invariant — progress no longer depends on how many `continue` paths the body has, which is the property that broke here.

`i` is not read after the loop, so incrementing before the `break` check is harmless. I checked the alias sequence against the original `for (int i = 0; true; i++)` semantics over every combination of 0–4 alias files × {absent, null, non-X509, non-matching, matching} anchors (341 cases): visit order and termination match in all of them.

`val alias = "$hash.${i++}"` is an equivalent one-liner if you prefer it; I went with the separate statement to keep the mutation out of the string template.

### Testing

No test is included. `isKnownRoot` reads `systemKeystore` and `systemCertificateDirectory`, both `private val`s with no injection point, and `KeyStore.getInstance("AndroidCAStore")` is unavailable off-device, so reaching this loop from a test would need either a seam in the object or an instrumented device with a disabled system CA. Happy to add one if you'd like it done a particular way.

I found this by code review while vendoring the file, not from a field report, so I have no device trace — the reachability argument above is the whole of the evidence.


djc (approved):
This makes sense to me but needs confirmation from someone more familiar with this code.

complexspaces (approved):
Nice catch here. I agree this looks like a correct fix.

ChoiSeungMyung:
This has two approvals — is there anything blocking the merge? Any sense of the timeline?

djc:
We're working on some other important issues with the Android setup:

- https://github.com/rustls/rustls-platform-verifier/issues/221

I figured it might make sense to merge this after we get CI passing again.

djc:
Okay, looks like CI failing was unrelated to that issue -- can you rebase on current `main`?

ChoiSeungMyung:
Rebased on current main — no conflicts. Thanks for the review!

djc:
Thanks for your patience!

PR #248, opened 2026-09-09
Generate Android network-security-config with CRL hosts

djc:


rustls-platform-verifier/tests/codegen.rs
```diff
@@ -0,0 +1,57 @@
...
+       https://github.com/rustls/rustls-platform-verifier/issues/221 -->
+  <domain-config cleartextTrafficPermitted=\"true\">
+";
+
+const FOOTER: &str = "\
+  </domain-config>
+</network-security-config>
+";
+
+const OUTPUT: &str = concat!(env!("CARGO_MANIFEST_DIR"), "/security-config.xml");
```
complexspaces: We can go ahead and write this to where the Android packaging will need it to be:


```suggestion

const OUTPUT: &str = concat!(env!("CARGO_MANIFEST_DIR"), "/android/rustls-platform-verifier/src/main/res/xml/security-config.xml");
```

rustls-platform-verifier/Cargo.toml
```diff
@@ -57,6 +57,9 @@ windows-sys = { version = ">=0.52.0, <0.62.0", default-features = false, feature
 
 [dev-dependencies]
 rustls = { version = "0.23", default-features = false, features = ["ring"] }
+similar-asserts = "1"
+tokio = { version = "1", features = ["macros", "rt-multi-thread"] }
```
complexspaces: Is the multi-threaded runtime needed? There's no parallelism AFAICT.

complexspaces (changes requested):

android-release-support/tests/codegen.rs
```diff
@@ -0,0 +1,61 @@
...
+            out,
+            "    <domain includeSubdomains=\"true\">{host}</domain>"
+        )
+        .unwrap();
+    }
+    out.push_str(FOOTER);
+
+    let stored = fs::read_to_string(OUTPUT).unwrap_or_default();
+    fs::write(OUTPUT, &out)?;
+    if stored != out {
```
complexspaces, line 34: As it looks like this can cause CI to start failing randomly (assuming a new CRL endpoint is stood-up), can we add some new documentation to [MAINTENANCE.md](https://github.com/rustls/rustls-platform-verifier/blob/main/admin/MAINTAINENCE.md) and tell the reader via `println!` to go there for a how-to on updating the list?

Cargo.lock
```diff
@@ -89,370 +246,1393 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
...
+checksum = "f87165f0995f63a9fbeea62b64d10b4d9d8e78ec6d7d51fb2125fda7bb36788f"
+
+[[package]]
+name = "rustls-webpki"
+version = "0.103.13"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "61c429a8649f110dddef65e2a5ad240f747e85f7758a6bccc7e5777bd33f756e"
+dependencies = [
+ "aws-lc-rs",
+ "ring",
```
complexspaces, line 1194: What's causing this to be pulled in still?

Cargo.lock
```diff
@@ -89,370 +246,1393 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
...
+ "rustls-platform-verifier-android 0.1.1",
+ "rustls-webpki",
+ "security-framework",
+ "security-framework-sys",
+ "webpki-root-certs",
+ "windows-sys 0.61.2",
+]
+
+[[package]]
+name = "rustls-platform-verifier"
```
complexspaces, line 1151: Do we need a crates.io patch definition to merge these together instead of building both the local repository and published crate for the test?

Cargo.lock
djc: It is... not clear? Probably some variant of the Cargo bug that keeps unactivated weak feature dependencies in the lockfile -- `cargo tree -i ring` says there is nothing to print.

android-release-support/tests/codegen.rs
djc: Added a hint to the test output and added a section in the maintenance file.

Cargo.lock
djc: That feels like a bunch of extra complexity with minimal benefit?

complexspaces: I'm fine with not doing this if its not required, I just ran into some annoyances with the same issue when pulling in `reqwest` to do some testing on Windows the other month where the changes made locally wouldn't be applied to the test's build tree.

djc: Suggest we leave it like this for now and change it when deemed necessary? The webpki-ccadb code is using `tls_certs_only()` on the reqwest client builder so I think rustls-platform-verifier isn't even used there.

complexspaces (approved):
This looks great to me, thanks for the help.

In a followup PR I will hookup the new `network_security_config.xml` to the Gradle packaging process and such. Its harmless to keep floating around between then though. With this in `main` I can do more accurate testing over in https://github.com/complexspaces/rustls-platform-verifier-android-sandbox anyway.

Issue #226, opened 2026-03-22
Docs for getting Kotlin/maven component seem incomplete


rnikander:
I followed the docs in the Readme to add to my build.gradle.kts script, and I get:

```
Execution failed for task ':app:checkDebugAarMetadata'.
> Could not resolve all files for configuration ':app:debugRuntimeClasspath'.
   > Could not resolve all dependencies for configuration ':app:debugRuntimeClasspath'.
      > Could not find any matches for rustls:rustls-platform-verifier:latest.release as no versions of rustls:rustls-platform-verifier are available.
        Searched in the following locations:
          - https://dl.google.com/dl/android/maven2/rustls/rustls-platform-verifier/maven-metadata.xml
          - https://repo.maven.apache.org/maven2/rustls/rustls-platform-verifier/maven-metadata.xml
          - file:/Users/rob/Workspace/external/rustls-platform-verifier/android-release-support/maven/rustls/rustls-platform-verifier/maven-metadata.xml
          - file:/Users/rob/Workspace/external/rustls-platform-verifier/android-release-support/maven/rustls/rustls-platform-verifier/
          - file:/Users/rob/Workspace/external/rustls-platform-verifier/android-release-support/maven/rustls/rustls-platform-verifier/.gitkeep/rustls-platform-verifier-.gitkeep.pom
          - file:/Users/rob/Workspace/external/rustls-platform-verifier/android-release-support/maven/rustls/rustls-platform-verifier/.gitkeep/rustls-platform-verifier-.gitkeep.jar
        Required by:
            project ':app'
```

That `android-release-support/` seems to be empty. The kotlin files are in `android/`. Am I supposed to compile something and copy it to `android-release-support/`?

Maybe this is because I have a local cloned repo. (I had to because I need to use JNI 0.22, which you have on a branch). 

Can I just copy the `.kt` files to my Android App project? That might be easier than dealing with Gradle.

complexspaces:
This should be fixed by https://github.com/rustls/rustls-platform-verifier/pull/251 now and the `0.7.1` release. As part of that I redid the documentation for the Maven component and the version resolution works much more normally now. Please let me know if the problem still persists!

PR #251, opened 2026-09-19
Migrate from Cargo-bundled Maven repository to GitHub-hosted Maven artifact repository

### LINKED ISSUE #115: Simplify Android release support by including a jar directly
The crate-local Maven repository didn't work for Signal because we use rustls-platform-verifier in a *library,* and our main app doesn't know about Rust at all. We determined that as long as no *other* library is using rustls-platform-verifier, and as long as rustls-platform-verifier's Java/Kotlin side doesn't need anything but its classes, [we can slurp those classes into our library directly](https://github.com/signalapp/libsignal/blob/a4a0663528dadc38215e46c6f94484b435f5fe02/java/android/build.gradle#L84-L113) and everything will be fine. (And indeed, it is working for us.)

This may have some downsides for general use (the main one being that `files(…)` dependencies can't have metadata), but given how minimal the crate-local Maven repository is, I'm not sure they'll come up in practice. So maybe this is a simpler approach for future versions of the crate?

```groovy
dependencies {
    implementation files(findRustlsPlatformVerifierClasses())
}

File findRustlsPlatformVerifierClasses() {
    def dependencyText = providers.exec {
        it.workingDir = new File("../")
        commandLine("cargo", "metadata", "--format-version", "1")
    }.standardOutput.asText.get()

    def dependencyJson = new JsonSlurper().parseText(dependencyText)
    def manifestFile = file(dependencyJson.packages.find { it.name == "rustls-platform-verifier-android" }.manifest_path)
    return new File(manifestFile.parentFile, "classes.jar")
}
```

complexspaces:
This PR implements the plan that I [outlined last year](https://github.com/rustls/rustls-platform-verifier/issues/115#issuecomment-2973361515) to redo the Android distribution channels (hopefully for the last time!) with a few improvements for simplicity.

We are switching from shipping AAR builds with Cargo releases to using `rustls-platform-verifier-android` as solely a version synchronization tool and distributing the AAR via a stateful Maven "local" repository that is going to be hosted in the `maven-archive` branch of this repository. Gradle will depend on this remote filesystem structure transparently, treating it exactly like a regular Maven package registry.

The `main` branch of this repository (and any release branches) now hold a `maven-metadata-local.xml` file that contains the persistent state required for a Maven remote. `main` is the source of truth for the data and during releases it gets copied into the `maven-archive` branch to "enable" finding the newly built artifacts. This is also intended to prevent ever running into Git conflicts when switching between the two.

This now means there shouldn't be anymore weird dependencies between the Rust part of this library and the Android part for downstream users who have a split between their Java libraries and actual host applications. Both can fetch from the same remote without any dependency on eachother and the final version of the Android component can be selected by the top-level app.

I tested this plan out in https://github.com/complexspaces/rustls-platform-verifier-android-sandbox originally with 1Password's Android app as the downstream user using the new README code snippets and all. Here's a screenshot of Android Studio showing successful version resolution and downloading of the AAR library:
image

More PRs follow this one to assemble the rest of the components and then demonstrate preparing for a new release using the updated process and releasing guide:
- #252 sets up the `maven-archive` branch that artifacts will be served on in the future. It shares the metadata XML file owned by this branch.
- #253 updates that Maven metadata file and Rust crate versions to setup for a new release that will be made including all these process changes. It demonstrates a bump to the checked in state.
- #254 pulls in the changes from both #252 and #253 to produce a newly updated Maven archive state that can serve the `0.1.2` release to Gradle clients over the network.

Commit-by-commit review is heavily recommended.

Closes #115, #226


djc (approved):

complexspaces:
Ah I forgot one thing last week: Before this merges I want to test reverting the test-only workaround for CRL fetching failures and confirm the new manifest is working as intended in our CI environment:
 https://github.com/rustls/rustls-platform-verifier/blob/027c029f1523ac8b75eca7996c72c2efb4b69b0b/android/rustls-platform-verifier/src/main/java/org/rustls/platformverifier/CertificateVerifier.kt#L360-L366

complexspaces:
I confirmed its working locally. With the inline workaround removed and changing the manifest locally to allow no cleartext traffic I get the same error everyone runs into:
image

Putting the manifest back to its `main` state allowing cleartext traffic for the CRL endpoints makes all tests pass 🎉 
image

I needed to bump our build tooling versions locally to work with the latest Android Studio as well so I've pushed that up here.

djc:
The commit history looks a bit messy? IMO the Kotlin formatting/lint stuff should be fixed in the same commit where the version that causes these to pop up is bumped.

It's not clear to me why CI against [48fc34f](https://github.com/rustls/rustls-platform-verifier/pull/251/commits/48fc34f2e4f3d1a1e3c235835569bc40cdd6dbcf) failed but fixing that by reordering commits might be nice, too.

djc (approved):

complexspaces:
Good point on the commit history 😅. I squashed the Kotlin lint fixes and workaround removal into larger ones since I think neither of those make sense on their own because they are dependent on a bigger change.

I also reordered history to have "fixing builds to make everything work" be before anything that changes Android Studio related things.

PR #252, opened 2026-09-19
Prepare maven-archive branch to host Maven registry and Android artifacts

complexspaces:
This PR sets up the newly made `maven-archive` branch (which was manually initialized based off my `android-maven-repo-migration` branch). I have removed every non-essential file from the branch to make its purpose more clear and avoid anything getting out of sync.

The Maven repository metadata is already prepared and reused from #251.

In addition, I've downloaded the `android-artifacts.zip` bundles from the past [0.1.0](https://github.com/rustls/rustls-platform-verifier/releases/tag/v%2F0.1.0) and [0.3.3](https://github.com/rustls/rustls-platform-verifier/releases/tag/v%2F0.3.3) releases and backfilled them into this branch as well to make the repository source migration smoother for downstream users and just more consistent overall. 

Please feel free to verify the provenance of the backported build archives by checking the SHA1 sums match everywhere like I intended.

PR #253, opened 2026-09-19
Prepare 0.2.0 release of Android components and 0.7.1

### LINKED ISSUE #221: Android verifier returns Revoked for Let's Encrypt certificates (CRL-only, no OCSP URL)

complexspaces:
This PR sets up for a new release that includes both the new CRL networking allowlist that is ready to ship and recent Android bugfixes. I've opted for a 0.2.0 release so that Cargo is forced to resolve a new version in downstream lockfiles, which will break existing Android/Gradle setups intentionally because there is no longer a Maven archive in the crates.io upload. The hope is that this will increase the speed at which users adopt the new Maven repository. It already has the old releases too, so there shouldn't be any reason _not_ to switch right away.

It also demonstrates the new Maven repository release process and shows that it works through all the new releasing guide steps.

~~This should not be merged until #251 is merged, at minimum, but maybe #252 too.~~

Closes #221 
Closes #179


djc (approved):

complexspaces:
`0.7.1` of `rustls-platform-verifier` and `0.2.0` of `rustls-platform-verifier-android` have been published to crates.io 🎉.

